PR #55, opened 2015-02-04
Update x11.xml

theatre-x:
added description


az0:
Gnome, KDE, XFCE are display managers, and X11 is a windowing system.

theatre-x:
Ah crap, I always get those mixed up.

PR #56, opened 2015-02-04
Update filezilla.xml

theatre-x:
updated copyright


az0:
Not needed per https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html

PR #60, opened 2015-02-04
Update gimp.xml

theatre-x:
added description


cleaners/gimp.xml
```diff
@@ -21,6 +21,7 @@
 -->
 <cleaner id="gimp">
   <label>GIMP</label>
+  <description>Photo editor</description>
```
az0: @theatre-x : More generally it is an image editor or graphics editor

PR #65, opened 2015-02-04
Update skype.xml

theatre-x:
added description

### DIFF
--- a/cleaners/skype.xml
+++ b/cleaners/skype.xml
@@ -21,6 +21,7 @@
 -->
 <cleaner id="skype">
   <label>Skype</label>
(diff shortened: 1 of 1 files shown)


cleaners/skype.xml
```diff
@@ -21,6 +21,7 @@
 -->
 <cleaner id="skype">
   <label>Skype</label>
+  <description>Video chat client</description>
```
az0: @theatre-x : "Chat client" is already translated and specific enough

az0:
I committed the 272e9d3 based on the comment above

Thank you for getting this rolling

PR #68, opened 2015-02-04
Update and rename xchat.xml to hexchat.xml

theatre-x:
updated name

### DIFF
--- a/cleaners/hexchat.xml
+++ b/cleaners/hexchat.xml
@@ -2,7 +2,7 @@
 <!--
 
     BleachBit
-    Copyright (C) 2014 Andrew Ziem
+    Copyright (C) 2015 Andrew Ziem
     http://bleachbit.sourceforge.net
 
     This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


Mailaender:
Might want to keep both. XChat may be dead, but in Linux distribution repositories it is still available.

ROCKNROLLKID:
I think we can merge this. He put in the description that it was formerly called XChat. 

Besides, a lot of the old Linux distros will be dropped when Python 3/GTK 3 will be done. New min version will be Ubuntu 14.04

cleaners/hexchat.xml
```diff
@@ -2,7 +2,7 @@
 <!--
 
     BleachBit
-    Copyright (C) 2014 Andrew Ziem
+    Copyright (C) 2015 Andrew Ziem
```
az0: This line is causing a merge conflict

az0:
This is just a name change, and it can still clean XChat files, so merging this is OK. The issue is the merge conflict.

ROCKNROLLKID:
I don't think theatre-x is around anymore, at least I have not seen him active in the BleachBit community in over a year. I can go ahead and open a new PR for this. Is the year suppose to be 2016 or 2008-2016?

az0:
The year can be just left off the change. I'll close this then.

PR #73, opened 2015-02-04
Update windows_defender.xml

theatre-x:
added description


az0:
I pulled apart the non-conflicting line and committed it: fe201fc

Thank you

PR #74, opened 2015-02-04
Update wine.xml

theatre-x:
added description

### DIFF
--- a/cleaners/wine.xml
+++ b/cleaners/wine.xml
@@ -2,7 +2,7 @@
 <!--
 
     BleachBit
-    Copyright (C) 2014 Andrew Ziem
+    Copyright (C) 2015 Andrew Ziem
     http://bleachbit.sourceforge.net
 
     This program is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


az0:
Two things.  By default only people who use Wine will see this cleaner, and people who use Wine should know what Wine is.  Adding the string also requires translation (63 languages and counting), so I would rather leave it off.  

I have used the description field for descriptions that can be shared (e.g., at least 10 cleaners share "Web browser").  In the future, maybe the GUI would organize them into groups.

Another good use of the description field is for something that people would have installed but are not sure what it is.

About the copyright, I recently learned that copyrights should not be just updated to the latest year (which I used to always do), and this explains it better: https://www.gnu.org/prep/maintain/html_node/Copyright-Notices.html   

Based on this, the copyright notice for Wine.xml would be "Copyright (C) 2009 Andrew Ziem" because that is the only year I made substantial changes to Wine.xml, and I am the only author.

theatre-x:
Understood with the copyright bit.

My reason to update and/or descriptions was because I noticed the lack of consistency with them i.e: some programs had a description and some didn't, and they also seemed, bluntly speaking, half-assed and felt it unprofessional.

I figure it would help people out if I went through and better elaborated, in a nutshell, what these cleaners are/do. :)

PR #75, opened 2015-02-04
Update winrar.xml

theatre-x:
added description


az0:
Please use "Archiver"  or "File archiver" consistent with these
http://www.rarlab.com/download.htm
https://en.wikipedia.org/wiki/WinRAR

az0:
Based on this conversation I committed a similar description 67c9a2fa80f04869a5c0217040a18bf0ee770af3

Thanks for the suggestion

PR #76, opened 2015-02-14
Update x11.xml

theatre-x:
made suggested change


az0:
Wikipedia calls it a "windowing system" http://en.wikipedia.org/wiki/X_Window_System

PR #81, opened 2015-02-14
Update deepscan.xml

theatre-x:
proposed description?

### DIFF
--- a/cleaners/deepscan.xml
+++ b/cleaners/deepscan.xml
@@ -21,6 +21,7 @@
 -->
 <cleaner id="deepscan">
   <label translate="true">Deep scan</label>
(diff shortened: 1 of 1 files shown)


tiemay:
Deep scan is somewhat vague, this description could be helpful.

az0:
I agree that there can be a better description, but in this case
>Deep clean backup and temporary files on the disk

"Deep clean" is similar to the cleaner name "deep scan," and "backup" and "temporary files" are not new information compared to the options.

Therefore, I suggest
>Clean files widely scattered across the disk

This gives some new information, and helps explain the warning "This option is slow."

az0:
Based on your suggestion I committed a similar description 

See 655c41d3d20ac518279fe31a5b67f92fd9f2115a

PR #86, opened 2015-03-14
added windows crashdumps cleaner

cfoellmann:
UNTESTED!!
havent had the time to look into how to compile, yet. @az0 can you test that one?

### DIFF
--- a/bleachbit/Cleaner.py
+++ b/bleachbit/Cleaner.py
@@ -482,6 +482,7 @@ class System(Cleaner):
             self.add_option('logs', _('Logs'), _('Delete the logs'))
             self.add_option(
                 'memory_dump', _('Memory dump'), _('Delete the file memory.dmp'))
+            self.add_option('crash_dumps', 'Crash dumps', _('Delete the crash dumps'))
             self.add_option('muicache', 'MUICache', _('Delete the cache'))
             # TRANSLATORS: Prefetch is Microsoft Windows jargon.
             self.add_option('prefetch', _('Prefetch'), _('Delete the cache'))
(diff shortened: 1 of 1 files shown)


ROCKNROLLKID:
There is a entry in Winapp2 that looks very similar to this:

[Memory Dumps]
ID=1015
LangSecRef=3021
FileKey1=%windir%|memory.dmp
FileKey2=%windir%\MiniDump|*.dmp
FileKey3=%LocalAppData%\CrashDumps|*.dmp

This was one of the entries taken out of Winapp2 and was re-added to BleachBit specifically. I suppose we can move it to CleanerML, but then I suggest you add the other FileKeys and then remove it from Winapp2.

cfoellmann:
What can we do here?
Move the OS (Windows) cleaners over to CleanerML? Or extend the core files?

ROCKNROLLKID:
Since it is no longer under Winapp2 care, I would suggest moving it to CleanerML.

cfoellmann:
Winapp2 was Never more than an option for bleachbit.
Currently the cleaners for win are in the bleachbit package - hardcoded.

ROCKNROLLKID:
Feel free to take any extra entries from the Winapp2 one and move it to your CleanerML and I will edit the Winapp2 file here on github and remove the entry out of there.

I think it should be done to all the entries that got removed from Winapp2 that we re-added in a modified copy of Winapp2. They should all be in CleanerML instead.

az0:
Yes, it looks like this could be a simple cleaner written in CleanerML. Because it affects so many Windows systems, though, I would like to merge it in this repository instead.

If the cleaner were named windows.xml, it would be a good place to move more of the cleaning options out of Python and into CleanerML.

In case you like to take to attempt it, I will leave this PR open.

ProactiveServices:
As someone who fixes Windows systems for a living I'd ask that you only take out dumps that are old, say at least a year old.

ROCKNROLLKID:
This pull request is not needed anymore. The removed entries have been moved back to the main Winapp2 respo and are managed by the community again. See here: https://github.com/MoscaDotTo/Winapp2/tree/master/Non-CCleaner

You can go ahead and close this.

az0:
Closing because of comment "This pull request is not needed anymore. "

PR #87, opened 2015-03-15
extend skype by old installers and patches

cfoellmann:


az0:
Is this related to automatic updates?

cfoellmann:
yes. the installed versions are not being removed after installation

PR #88, opened 2015-03-19
Replace hardcoded localization paths with CleanerML-entries

tstenner:
This is the same as the last pull request but based on the current master so there's no merge conflict

### DIFF
--- a/bleachbit/Cleaner.py
+++ b/bleachbit/Cleaner.py
@@ -576,6 +576,9 @@ class System(Cleaner):
         # unwanted locales
         if 'posix' == os.name and 'localizations' == option_id:
             for path in Unix.locales.localization_paths(locales_to_keep=options.get_languages()):
+                if os.path.isdir(path):
+                    for f in FileUtilities.children_in_directory(path, True):
+                        yield Command.Delete(f)
                 yield Command.Delete(path)
 
         # Windows logs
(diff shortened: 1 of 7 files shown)


az0:
@tstenner   This looks interesting and I would like to take a closer look.  I did notice that the Travis CI build failed.  Please see https://travis-ci.org/az0/bleachbit/jobs/55505346#L1438

tstenner:
The test passes on my machine, I'm assuming that the VMs has linked /usr/share/locale somewhere else. After removing the check if the path starts with `/usr/share/locale` everything passes.

tstenner:
I noticed that directories aren't deleted recursively at the moment. What would be the best way to implement this (e.g. as flag to `Delete(path,recurse=True)`)?

az0:
To delete recursively, try something like "for f in FileUtilities.children_in_directories(path, True): delete(path)" to iterate through all the children in a directory.  Then remember to delete the parent.

tstenner:
Thanks for your suggestion, I already implemented it and it's working on my machine (still waiting for Travis).

az0:
Travis CI looks OK (except for an apparently-unrelated network error).  Please allow some patience for me to look more at the whole thing.  Thank you.

az0:
Released in BleachBit 1.7.7 beta.  Please test:
http://bleachbit.sourceforge.net/news/bleachbit-177-beta

PR #92, opened 2015-06-20
Some further localization changes

tstenner:
I added some two letter codes from https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes and added the localization paths for MuseScore 2, manpages and qt4.


az0:
Thanks!

az0:
@tstenner : Do you mind looking at these two issues related to localizations?

1. "Not delete some languages folders" https://bugs.launchpad.net/bleachbit/+bug/1471883
2. "debug: '/usr/share/bleachbit/cleaners/localizations.xml' is not usable http://bleachbit.sourceforge.net/node/1145#comment-1812

I usually manage bug tickets in Launchpad, but if you do not mind and if it would help, I could assign these to you in GitHub or Launchpad.

tstenner:
The first bug should be mostly resolved when this commit appears in a release (was merged 11 hours after the 1.8 release).
The second bug is quite harmless: as localizations.xml contains no <action>, the debug message is generated. I'll see what I can do about that.

PR #94, opened 2015-07-01
Fix preferences dialog resize.

abobov:
Label/buttons stay fixed at top/bottom and list control take all space.


az0:
I think you fixed this [old bug](https://bugs.launchpad.net/bleachbit/+bug/1004278).  Thank you!  

Please note there is a branch for GTK+ 3, and it is on my to-do list to update it with the changes from master.

PR #95, opened 2015-07-07
Resolve "localizations.xml is not usable"

tstenner:
This adds a dummy action to prevent the warning about localizations.xml not being usable.
Resolves http://bleachbit.sourceforge.net/node/1145


az0:
Thank you!

PR #99, opened 2015-08-05
Localizations simplification

tstenner:
Since the Qt localizations for Qt programs are in separate folders, I added support for regular expressions in filters via the new `<regexfilter>`-tag.
The previous filters (`<path location="." filter="foo*.bar"/>`) are automatically rewritten to regular expressions (here equivalent to `<regexfilter prefix="foo" postfix="\.bar"/>`), so existing filters don't need to be changed.

I already added some examples with comments in localizations.xml, someone with a bunch of Qt applications installed should contribute some more filters.

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -282,14 +282,14 @@ class Locales:
         if not os.path.isdir(path):
             return
 
-        pre = ''
-        post = ''
+        pre = None
+        post = None
         if 'regexfilter' == xmldata.nodeName:
             pre = xmldata.getAttribute('prefix') or ''
             post = xmldata.getAttribute('postfix') or ''
         if 'path' == xmldata.nodeName:
             userfilter = xmldata.getAttribute('filter')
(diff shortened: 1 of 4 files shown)


tstenner:
One test fails, because `assertIsNone` was added to unittest in python 2.7. Should I remove it?

az0:
Would you please switch it to something like assertTrue(x is None) ? I want to depreciate older Python versions but have not done it yet.

Thank you for this PR and I hope to take a closer look soon

tstenner:
I added fixes for a regression concerning filters in the form of `<path location="/usr/share/locale"/>` without an explicit filter or `filter="*"` and made the filter for Qt translations a little broader.

az0:
I appreciate the submission and need more time to give it a closer look. Thank you for your patience

az0:
Closing and re-opening to test coveralls.io

az0:
Reopening to test coveralls.io

PR #101, opened 2015-10-03
Additional 3-letter language codes

tstenner:
This adds all ISO-639-3 codes for translations I have on my installation, freeing an additional 10.3MB.


az0:
Nice. Thank you

Issue #40, opened 2014-12-07
update Windows installer for GTK+ 3


az0:
@moimael :  I am guessing you do not want to work on this one.

az0:
My next step is to try py2exe/cx_freeze on a  "hello world" GTK+ 3 app.

az0:
As an exercise, I froze a ["hello world" GTK+ v3 and Python 2.7 app for Windows](https://github.com/az0/frozen_pgtk3), and it is 93MB. The next step is cutting back the size. The directory ```build/exe.win32-2.7/share/locale``` alone is 48MB and looks like a good place to start.

ROCKNROLLKID:
Just curious, how far along is GTK3 to be added to BleachBit? Will we see it in 2.0 or will it take a little longer?

az0:
I haven't had time yet to shrink the Windows installer to a manageable size. It will probably not happen for version 2.0

ROCKNROLLKID:
@az0 Maybe you can try this to help shrink the installer: http://wixtoolset.org/

az0:
I used WiX a long time ago, and while it would be interesting to switch from a .exe to .msi installer, the issue isn't so much how the files are bundled (like, using NSIS, WiX, InnoSetup, etc.). The issue is that before being packed up, they are huge because of the giant GTK+ version 3 runtime, so unnecessary files need to be deleted (like I do today with the GTK+ 2 runtime). If that is not enough, it may be necessary to alter files (like I do now also with some GTK+  2 files), or recompile GTK+ 3 to remove unneeded code, or install GTK+ 3 separately. Each of these has its costs and tradeoffs.

az0:
This was implemented. For example, here are some GTK3+ Windows builds (alpha level)

https://ci.appveyor.com/project/az0/bleachbit/build/2.0.682/artifacts

PR #108, opened 2015-10-14
Add size column to side tree control.

ihewitt:
This adds a size column to the left hand tree control making it easier to see where space is being used.
I simply added this as a third column after the id column rather than inserting it to reduce the number of changes, if preferred I can reorder the columns.

### DIFF
--- a/bleachbit/CLI.py
+++ b/bleachbit/CLI.py
@@ -58,6 +58,10 @@ class CliCallback:
         """Not used"""
         pass
 
+    def update_item_size(self, op, opid, size):
+        """Not used"""
+        pass
+
     def worker_done(self, worker, really_delete):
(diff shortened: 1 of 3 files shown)


az0:
This is an interesting and efficient solution that works well in my initial testing. Thank you.

Please note I am avoiding many changes to the GUI in the master branch. It uses GTK+ version 2, while there is a modernization branch for GTK+ version 3.  However, the branch has not been updated (using rebase) in a while while I try to figure out issue #40, so when that branch is updated, I may ask you to reapply this change there.

ihewitt:
Great thanks, yes I wanted to keep it really simple. I was considering what would be involved creating a grid on the right hand so I'll think about that for the gtk3 version.

Issue #103, opened 2015-10-03
Use binary-prefix file sizes


Wuzzy2:
LOL, I just noticed that BleachBit calculates file sizes in factors of 1000 (1 BleachBit Kilobyte = 1000 Bytes), rather than in factors of 1024. While this is stricly SI-conform, it is very confusing because many operatings systems and programs use 1024 as factor.

The best solution would be to simply switch to binary prefixes:

1 KiB = 1024 B, 1 MiB = 1024 KiB, and so on.

See here: https://en.wikipedia.org/wiki/Binary_prefix#Adoption_by_IEC.2C_NIST_and_ISO

az0:
Since March 2010 BleachBit should be following this policy.

https://wiki.ubuntu.com/UnitsPolicy

If you still think this is an issue, please reopen

Wuzzy2:
I am unable to reopen this issue, but I can still comment on this.

I call bullshit on Ubuntu's UnitsPolicy. It does not account for the fact that the past 25 years or so used factor 1024 for “fake SI” style prefixes (KB, MB, GB, etc.). It is a rather young interpretation that the “correct” factor is 1000, because of SI. No matter in which camp you are, this basically means that there are two competing interpretations of what “1 MB” means, one uses factor 1024 and is far more wide-spread, while the other uses factor 1000. This creates an ambiguity. Whenever you read “1 MB” you can't trust this number unless you know which factor the author has been using.

I still propose to switch to binary-style prefixes (KiB, MiB, GiB, etc.), since this preserves the factor 1024 which has been used for so long and resolves the ambiguity.

tiemay:
Perhaps a compromise?
> For file sizes there are two possibilities:

> 1. Show both, base-10 and base-2 (in this order). An example is the Linux kernel: "2930277168 512-byte hardware sectors: (1.50 TB/1.36 TiB)"
> 1. Only show base-10, or give the user the opportunity to decide between base-10 and base-2 (the default must be base-10). 

Maybe we could do one of those two options?

Also, the KDE system settings lets you choose between IEC and Metric, and most KDE programs abide by that setting. I'm not sure if a GTK application could make use of that or not, but it would be nice.

tstenner:
Is anyone working on this? If no, should this be implemented as "show both" or "make it an option"?

az0:
I am not working on this.

az0:
Fixed by @tstenner's pull request #111

PR #111, opened 2015-12-14
Add option for preferred size prefixes (#103)

tstenner:
This adds an option to switch between IEC and SI prefixes (see #103).
It also fixes the case where no value was returned if more than 1 petabytes were freed.

### DIFF
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -92,31 +92,39 @@ def __random_string(length):
 
 
 def bytes_to_human(bytes_i):
-    """Display a file size in human terms (megabytes, etc.) using SI standard"""
+    """Display a file size in human terms (megabytes, etc.) using preferred standard (SI or IEC)"""
 
-    storage_multipliers = {1000 ** 5: 'PB', 1000 ** 4: 'TB',
-                           1000 ** 3: 'GB', 1000 ** 2: 'MB', 1000: 'kB', 1: 'B'}
+    if bytes_i < 0:
+        return '-' + bytes_to_human(bytes_i)
+
+    from Options import options
+    if options.get('units_iec'):
+        prefixes = ['','Ki','Mi','Gi','Ti','Pi']
+        base = 1024.0
+    else:
+        prefixes = ['','k','M','G','T','P']
+        base = 1000.0
 
     assert(isinstance(bytes_i, (int, long)))
 
     if 0 == bytes_i:
         return "0"
 
-    if bytes_i >= 1000 ** 3:
+    if bytes_i >= base ** 3:
         decimals = 2
-    elif bytes_i >= 1000:
+    elif bytes_i >= base:
         decimals = 1
     else:
         decimals = 0
 
-    for key in sorted(storage_multipliers.keys(), reverse=True):
-        if bytes_i >= key:
-            abbrev = round((1.0 * bytes_i) / key, decimals)
-            suf = storage_multipliers[key]
-            return locale.str(abbrev) + suf
-
-    if bytes_i < 0:
-        return "-" + bytes_to_human(abs(bytes_i))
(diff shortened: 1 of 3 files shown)


az0:
Interesting. Thank you. I will take a closer look. For now, there is a recursion error according to the Travis and Appveyor

tstenner:
Ah, missed the a sign in case of negative amounts.

bleachbit/GuiPreferences.py
```diff
@@ -178,6 +178,11 @@ def __general_page(self):
             'toggled', self.__toggle_callback, 'delete_confirmation')
         vbox.pack_start(cb_popup, False)
 
+        # Use base 1000 oder 1024?
+        cb_units_iec = gtk.CheckButton(_("Use IEC sizes (1 KiB = 1024 bytes) instead of SI (1 kB = 1000 bytes)"))
```
az0: I know so far all the other options in this menu use CheckButton, but maybe this choice would be more easier for the user to understand this choice as a radio button? Check boxes are usually used to indicate something is enabled or disabled, while radio boxes are a set of mutually-exclusive options. Conceivably there could be a third option other than SI or IEC. Or another way of thinking about it that "IEC enabled and SI disabled" could be "SI enabled and IEC disabled," so the enabled/disabled is not intuitive.

bleachbit/FileUtilities.py
```diff
@@ -92,31 +92,39 @@ def __random_string(length):
...
-    """Display a file size in human terms (megabytes, etc.) using SI standard"""
+    """Display a file size in human terms (megabytes, etc.) using preferred standard (SI or IEC)"""
 
-    storage_multipliers = {1000 ** 5: 'PB', 1000 ** 4: 'TB',
-                           1000 ** 3: 'GB', 1000 ** 2: 'MB', 1000: 'kB', 1: 'B'}
+    if bytes_i < 0:
+        return '-' + bytes_to_human(-bytes_i)
+
+    from Options import options
+    if options.get('units_iec'):
```
az0: Would you please add at least a basic test to tests/TestFileUtilities.py? It doesn't have to cover the round trip. It could be a new test function just for IEC.

bleachbit/FileUtilities.py
```diff
@@ -92,31 +92,39 @@ def __random_string(length):
...
+        prefixes = ['','Ki','Mi','Gi','Ti','Pi']
+        base = 1024.0
+    else:
```
az0: Because this function can be called a thousand times (once for each file, like when cleaning localizations), I ran this benchmark in three scenarios
````timeit.timeit('bleachbit.FileUtilities.bytes_to_human(100000)', setup='import bleachbit.FileUtilities', number=100000)````

Original code: 4.7 seconds
New code: 9.2 seconds
New code that replaces conditional with two global variables: 4.2 seconds

And just this one new line takes 3.5 seconds
````timeit.timeit('options.get("units_iec")',setup='from bleachbit.Options import options', number=100000)````

However, bytes_to_human on 1000 files, which is more typical than 100,000 files, it is only 0.1 seconds, so maybe not worth optimizing.

bleachbit/GuiPreferences.py
tstenner: I also thought of using a ComboBox, but it's quite cumbersome without `gtk.combo_box_new_text()` (which is deprecated in PyGTK 2.24) and the replacement `gtk.ComboBoxText` was added in 2.24.
If it's ok for you, I'd have a go at it after the transition to GTK 3.

bleachbit/FileUtilities.py
tstenner: On the other hand 4 seconds is quite short compared to deleting 100K files and printing 100K lines.

az0: Yes

PR #113, opened 2015-12-22
Don't add a dummy option to localizations.xml (fixes #110)

### LINKED ISSUE #110: Exception while running operation 'localizations'
Traceback in BleachBit 1.9.2 beta on Slackware 14.1 x86_64 
````
File "/usr/share/bleachbit/Worker.py", line 330, in run_operations
for dummy in self.clean_operation(operation):
File "/usr/share/bleachbit/Worker.py", line 127, in clean_operation
assert(isinstance(operation_options, list))
AssertionError
```
[Strange GUI](http://i.imgur.com/nwAUP8R.jpg) where Localizations is at the top level (not under System)

Reported by [BBUser on the forum](http://bleachbit.sourceforge.net/node/1197)

@tstenner Do you think this is related to changes to put localizations in XML? Would you please mind taking a closer look?

tstenner:
With `auto_hide` disabled, the dummy action is shown in the menu but doesn't actually work.

### DIFF
--- a/bleachbit/CleanerML.py
+++ b/bleachbit/CleanerML.py
@@ -180,8 +180,6 @@ class CleanerML:
         for localization_node in localization_nodes:
             for child_node in localization_node.childNodes:
                 Unix.locales.add_xml(child_node)
-        # Add a dummy action so the file isn't reported as unusable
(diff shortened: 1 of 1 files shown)


az0:
thank you but I found an alternative way

PR #122, opened 2016-02-18
SQL and Mac OSx Changes.

brahmastra2016:
Zeroblob SQL query unique constrain fix.
Mac OSx Support added.
     a) Diagnostic "--sysinfo" knob work both in GUI and CLI
     b) Cleans:
                    Rotated Logs.
                    Firefox.
                    Deep Scan (Delete DS_Store)
                   Recycle Bin


bleachbit/Special.py
```diff
@@ -41,12 +41,12 @@ def __shred_sqlite_char_columns(table, cols=None, where=""):
     """Create an SQL command to shred character columns"""
     cmd = ""
     if cols and options.get('shred'):
+        #cmd += "update %s set %s %s;" % \
```
az0: I would like to address the SQL error in another way, so would you please remove it from this pull request?

brahmastra2016:
i had tried that approach 'if col !="url"' but then it would fail with the same  unique constrain error at 
"update moz_host = zeroblob(lenght(host))"

az0:
I thought I suggested a different approach.

brahmastra2016:
sorry i read my own code and got confused.

i change my pull request for now.

just a update .....
i saw that Firefox database (place.sqlite) changes and URL is no more set to UNIQUE 
now the problem is just host in moz_hosts

screen shot 2016-02-19 at 6 32 50 pm

brahmastra2016:
Will Reopen new with other Mac OSx Changes

PR #123, opened 2016-02-20
Mac OSx - Update 

brahmastra2016:
Mac OSx Support added.
1) Diagnostic "--sysinfo" knob work both in GUI and CLI
2) Cleans:
3) Rotated Logs.
4) Firefox.
6) Deep Scan
5) Recycle Bin
6) Safari.xml - Added Mac OSx Group
7) Added Processing running check for Mac OSx.

 Have first ever mac complied DMG of Bleachbit , like to share it with the community for alpha testing.

[bleachbit.dmg.zip](https://github.com/az0/bleachbit/files/139392/bleachbit.dmg.zip)


az0:
Thanks. I will review.

bleachbit/Special.py
```diff
@@ -42,10 +42,10 @@ def __shred_sqlite_char_columns(table, cols=None, where=""):
     cmd = ""
     if cols and options.get('shred'):
         cmd += "update %s set %s %s;" % \
-            (table, ",".join(["%s = randomblob(length(%s))" % (col, col)
```
az0: This is not related to Mac OS X, and I would like to fix it a different way, please.

cleaners/safari.xml
```diff
@@ -19,36 +19,64 @@
...
+    </group>
   </option>
   <option id="history">
     <label>History</label>
     <description>List of visited web pages</description>
-    <action command="delete" search="file" path="$APPDATA\Apple Computer\Safari\History.plist"/>
-    <action command="delete" search="file" path="$APPDATA\Apple Computer\Safari\LastSession.plist"/>
-    <action command="delete" search="walk.files" path="$localappdata\Apple Computer\Safari\History\"/>
-    <action command="delete" search="walk.files" path="$localappdata\Apple Computer\Safari\Webpage Previews\"/>
+    <group os="windows">
```
az0: `<group>` is not supported in the XSD or the Cleaner.py code. What are you trying to do? When running on one operating system, the files from the other OS will simply not exist and be ignored. If you just want to organize, please use XML comments instead.

Also in the future I plan to improve this by defining cleaner-level variables that would look something like this
````
<var os="darwin" name="safari" value="~/Library/Safari/" />
<var os="windows" name="safari" value="%APPDATA%\Apple Computer\Safari\" />
````
Then each action, like cleaning History.plist, can reference this variable.

bleachbit/Special.py
brahmastra2016: I am not sure what is the ask here.i have already revert back to zeroblob and randomblob.

as I said in one of previous post , firefox has already remove the unique contrain on url and we dont see the sql error anymore both on mac and windows.

thank you :)


Sent via the Samsung Galaxy Note® 3, an AT&T 4G LTE smartphone

<div>-------- Original message --------</div><div>From: Andrew Ziem <notifications@github.com> </div><div>Date:03/09/2016  11:57 PM  (GMT-05:00) </div><div>To: az0/bleachbit <bleachbit@noreply.github.com> </div><div>Cc: brahmastra2016 <Brahmastra2016@gmail.com> </div><div>Subject: Re: [bleachbit] Mac OSx - Update  (#123) </div><div>
</div>In bleachbit/Special.py:

> @@ -42,10 +42,10 @@ def __shred_sqlite_char_columns(table, cols=None, where=""):
>      cmd = ""
>      if cols and options.get('shred'):
>          cmd += "update %s set %s %s;" % \
> -            (table, ",".join(["%s = randomblob(length(%s))" % (col, col)
This is not related to Mac OS X, and I would like to fix it a different way, please.

PR #128, opened 2016-03-06
Preserve Datastore transactions, Prefetch

ProactiveServices:
SoftwareDistribution\Datastore .log files are database transaction logs, not text log files. These should not be deleted as it will damage Windows Update.
Removal of Prefetch files results in performance degradation. Prefetch is designed to improve performance so it should be left alone.

### DIFF
--- a/bleachbit/Cleaner.py
+++ b/bleachbit/Cleaner.py
@@ -607,7 +607,6 @@ class System(Cleaner):
                 '$windir\\security\\logs\\*.log',
                 '$windir\\security\\logs\\*.old',
                 '$windir\\SoftwareDistribution\\*.log',
-                '$windir\\SoftwareDistribution\\DataStore\\Logs\\*',
                 '$windir\\system32\\TZLog.log',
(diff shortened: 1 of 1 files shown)


az0:
Thank you for the PR.

SoftwareDistribution\Datastore: For people who do need Windows Updates history, cache, or rollbacks, I have heard it is safe to delete the contents of the folder. In what way do you mean "damage"?

Like prefetch files, browser cache also is designed to improve performance, so performance is not a sufficient reason to stop deleting prefetch. 

Please provide more information or revise the PR.

ProactiveServices:
The "Download" subfolder is safe to delete. This is just the update installers and can often be pretty big. Nothing inside is required to rollback etc. IME, Windows Update doesn't even seem to bother cleaning it up!

The DataStore holds a database, and associated metadata, of which updates are installed, data about update dependencies, installed Windows components, which updates have been superceded etc. for all Windows components. This is why the .edb is so big. You can delete DataStore (or the entire SoftwareDistribution folder) and Windows Update will re-create it all, but it will take an awful long time these days.

However - importantly - if you break the database it can prevent updates from being detected or installing correctly as it is left in an inconsistent state.

Re: Prefetch - what's the perceived benefit from removing it?

PR #129, opened 2016-03-11
Mac OSx - Update

brahmastra2016:
Mac OSx Support added.
1) Diagnostic "--sysinfo" knob work both in GUI and CLI
2) Cleans:
3) Rotated Logs.
4) Firefox.
6) Deep Scan
5) Recycle Bin
6) Safari.xml - Added Mac OSx Group
7) Added Processing running check for Mac OSx.

Have first ever mac complied DMG of Bleachbit , like to share it with the community for alpha testing.


bleachbit/Special.py
```diff
@@ -42,10 +42,10 @@ def __shred_sqlite_char_columns(table, cols=None, where=""):
     cmd = ""
     if cols and options.get('shred'):
         cmd += "update %s set %s %s;" % \
-            (table, ",".join(["%s = randomblob(length(%s))" % (col, col)
```
az0: This file has a merge conflict with commit c43077dcf55d79f20706da66d1aaacd6e800ee87

cleaners/safari.xml
```diff
@@ -19,36 +19,48 @@
...
+      <action command="delete" search="file" path="$APPDATA\Apple Computer\Safari\Cookies\Cookies.plist"/>
+      <action command="delete" search="walk.files" path="$localappdata\Apple Computer\Safari\LocalStorage\"/>
+      <action command="delete" search="file" path="~/Library/Safari/Cookies.plist"/>
+      <action command="delete" search="walk.files" path="~/Library/Safari/LocalStorage/"/>
   </option>
   <option id="history">
     <label>History</label>
     <description>List of visited web pages</description>
-    <action command="delete" search="file" path="$APPDATA\Apple Computer\Safari\History.plist"/>
-    <action command="delete" search="file" path="$APPDATA\Apple Computer\Safari\LastSession.plist"/>
```
az0: It seems you changed the indentation of lines that otherwise do not need to be touched. :) While this may seem picky, it unnecessarily obscures the code history.

To avoid this, there is a command "make pretty" that formats the XML files with standard indentation.

bleachbit/Cleaner.py
```diff
@@ -216,7 +216,10 @@ def __init__(self):
         self.add_option('vacuum', _('Vacuum'), _(
             'Clean database fragmentation to reduce space and improve speed without removing any data'))
 
-        if 'posix' == os.name:
+        if  'posix' == os.name and sys.platform.startswith('darwin'):
```
az0: Good

bleachbit/Cleaner.py
```diff
@@ -497,7 +501,8 @@ def __init__(self):
             # the uninstallers for software updates.
             self.add_option('updates', _('Update uninstallers'), _(
                 'Delete uninstallers for Microsoft updates including hotfixes, service packs, and Internet Explorer updates'))
-
+        elif 'posix' == os.name and sys.platform.startswith('darwin'):
```
az0: While this adds the option (the checkbox), I do not see a corresponding code to actually clean the recycle bin

bleachbit/Common.py
```diff
@@ -107,6 +107,10 @@
 # system cleaners
 if sys.platform.startswith('linux'):
     system_cleaners_dir = '/usr/share/bleachbit/cleaners'
+
+if sys.platform.startswith('darwin'):
```
az0: To avoid repeating the same directory, you could also use an `or` condition after `startswith(`linux`)`

bleachbit/Common.py
```diff
@@ -139,6 +143,8 @@
     # system-wide installed locale
     if sys.platform.startswith('linux'):
         locale_dir = "/usr/share/locale/"
+    elif sys.platform.startswith('darwin'):
```
az0: To avoid repeating the same directory, you could also use an `or` condition after `startswith(`linux`)`

bleachbit/Cleaner.py
brahmastra2016: it have update the recycle bin location for mac and it does work fine.as
you can see in attached pic.

az0: The attached picture does not make it. I did a search for `recycle_bin` and there is no Mac-specific code (besides `add_option`).

ROCKNROLLKID:
Just letting you know that maybe now it might not seem like people are interested in this because BleachBit has never had a Mac version up, but there has been quite a bit of users suggesting Mac versions throughout the years on the BleachBit forums. I am sure once this is done and you give it some time, lots of people will be using this on Mac and the interest will spike up.

You have come so far that it would be a shame to back down now.

brahmastra2016:
will start adding in small step with new pull request.

PR #130, opened 2016-03-20
Added ~/.kde4 folders everywhere.

Mailaender:
On @openSUSE everything is placed in `~/.kde4`. This is a rebase of https://build.opensuse.org/package/view_file/openSUSE:Factory/bleachbit/bleachbit-kde-patch.patch?expand=1 from version 1.6

### DIFF
--- a/cleaners/kde.xml
+++ b/cleaners/kde.xml
@@ -26,13 +26,17 @@
     <label>Cache</label>
     <description>Delete the cache</description>
     <action command="delete" search="glob" path="~/.kde/cache-*/*/*"/>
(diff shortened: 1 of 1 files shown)


az0:
I like how you did not remove the ~/.kde/, which is still used by some modern systems or older versions

Mailaender:
Yeah, I am not sure. It could be used by KDE3 or the `~/.kde4` could be a SUSE customization as it is one of the few distributions that let's you still install https://build.opensuse.org/project/show/KDE:KDE3 side by side and every other distribution uses `~/.kde` for version 4 but `¯\_(ツ)_/¯` and keeping it doesn't hurt.

PR #142, opened 2016-09-10
Added an appdata.xml file

Mailaender:
This gets you listed in Linux app stores. https://people.freedesktop.org/~hughsient/appdata/ has the details on this file. I took the screenshot from http://www.bleachbit.org/screenshots but it seems you moved to some kind of cache so I am not sure if these will stay forever. You may want to replace the URL.


az0:
Thank you!

Closes https://bugs.launchpad.net/bleachbit/+bug/1322911

PR #144, opened 2016-10-01
localizations.xml: Allow patterns in foldernames

tstenner:
Many Qt programs store their localizations in `/usr/share/XY/translations/XY_en.qm` instead of in `/usr/share/locale/en/XY`.
This PR adds a new attribute `pattern` to `<path>` elements which recurses into all matching folders.

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -39,11 +39,68 @@ import FileUtilities
 import General
 
 
-class Locales:
+class LocaleCleanerPath:
+    """This represents a path with either a specific folder name or a folder name pattern.
+    It also may contain several compiled regex patterns for localization items (folders or files)
+    and additional LocaleCleanerPaths that get traversed when asked to supply a list of localization
+    items"""
 
+    def __init__(self, location):
+        if location is None:
+            raise RuntimeError("location is none")
+        self.pattern = location
+        self.children = []
+
+    def add_child(self, child):
+        """Adds a child LocaleCleanerPath"""
+        self.children.append(child)
+        return child
+
+    def add_path_filter(self, pre, post):
+        """Adds a filter consisting of a prefix and a postfix
+        (e.g. 'foobar_' and '\.qm' to match 'foobar_en_US.utf-8.qm)"""
+        try:
+            regex = re.compile('^' + pre + Locales.localepattern + post + '$')
+        except Exception as errormsg:
+            raise RuntimeError("Malformed regex '%s' or '%s': %s" % (pre, post, errormsg))
+        self.add_child(regex)
+
+    def get_subpaths(self, basepath):
+        """Returns direct subpaths for this object, i.e. either the named subfolder or all
+        subfolders matching the pattern"""
+        if isinstance(self.pattern, re._pattern_type):
+            return (os.path.join(basepath, p) for p in os.listdir(basepath)
+                    if self.pattern.match(p) and os.path.isdir(os.path.join(basepath, p)))
+        else:
+            path = os.path.join(basepath, self.pattern)
+            return [path] if os.path.isdir(path) else []
+
+    def get_localizations(self, basepath):
+        """Returns all localization items for this object and all descendant objects"""
+        for path in self.get_subpaths(basepath):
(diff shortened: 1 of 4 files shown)


az0:
This looks helpful. Thank you.

Would you please add a unit test that creates a temporary directory, cleans it using pattern, and then validates it was deleted? There are several cases that test like this, for example see test_fake() in TestWinapp2.py

tstenner:
I'll see what I can do. Currently I'm a bit unhappy that the xml tree has to be walked for each directory entry when there's a pattern present, so I'll change that first.
(This should make the test a bit easier to implement as well)

az0:
OK, I will stand by. Thank you

tstenner:
Is there a document with style conventions (e.g. PEP-8, all lines shorter than 80 chars) and compatibility requirements (Python 2.6 and newer)? If Python 2.6 isn't required any more, I'd try to make it compatible with Python 3 as well.

az0:
Good question. Please see the new CONTRIBUTING.md.

While supporting Python 2 and 3 at once is OK, I do not plan to fully support Python 3 until I tackle GTK 3, which is in the modernization branch. It is not ready because of the issues documented here in GitHub.

bleachbit/Unix.py
```diff
@@ -39,11 +39,68 @@
...
+class Locales:
     """Find languages and localization files"""
 
-    localepattern = r"(?P<locale>[a-z]{2,3})(?:[_-][a-zA-Z]{2,4})?(?:\.[a-zA-Z0-9-]*)?(?:@[a-zA-Z]*)?"
+    # The regular expression to match locale strings and extract the langcode.
+    # See test_locale_regex() in tests/TestUnix.py for examples
+    # This doesn't match all possible valid locale strings to avoid
+    # matching filenames you might want to keep, e.g. the regex
+    # to match jp.eucJP might also match jp.importantfileextension
+    localepattern =\
```
tstenner: I'm not sure how specific the regex should be. it shouldn't match "en.extension", but all possible encodings would be even more complicated.

tests/TestUnix.py
```diff
@@ -96,16 +97,17 @@ def test_locale_regex(self):
                  ('en_US', 'en'),
                  ('en_US@piglatin', 'en'),
                  ('en_US.utf8', 'en'),
+                 ('ko_KR.eucKR', 'ko'),
                  ('pl.ISO8859-2', 'pl'),
-                 ('sr_Latn', 'sr'),
```
tstenner: This isn't matched any more, see the comment on the changed regex.

tests/TestUnix.py
```diff
@@ -121,7 +123,55 @@ def test_localization_paths(self):
             self.assert_(path.find('/en_') == -1)
             counter += 1
         self.assert_(
-            counter > 0, 'Zero files deleted by localization cleaner.  This may be an error unless you really deleted all the files.')
+            counter > 0, 'Zero files deleted by localization cleaner.'
```
tstenner: Line plit to conform with pep 8

tstenner:
This tests most of the code, I'll see if I can increase the coverage even more.
Is there any special reason to support Python 2.5? As far as I can see it the tests aren't run for it and even Debian oldstable has 2.7

ROCKNROLLKID:
There are still some builds for older Fedora, Ubuntu, and OpenSUSE users, but those will probably eventually get dropped anyways, so Python 2.6/2.7 should be enough.

tstenner:
I think Debian oldstable and the previous Ubuntu LTS already have 2.7, so the last active distribution might be windows.
When everything's ready, I'll rebase the commits and open a new PR so the history will be cleaner

ROCKNROLLKID:
Maybe we could consider dropping 2.6, as well and just sticking with 2.7. It would help clean up the code some bit, too.

az0:
Unix.py and Linux systems in general do not need to support Python 2.5, but the Windows build uses 2.5 because it makes the .exe installer smaller. At some point in the not to distant future I will update the Windows build to Python 2.7.

Thanks for the continued improvements, and I'll wait for the rebase.

PR #146, opened 2016-10-04
Ensure compatibility for non-ascii usernames

marcolagi:
Summary of changes:

- Defined a new constant in `Common.py`, `FSE`, containing the filesystem encoding
- Encoded/Decoded a bunch of basestrings using this constant in `Action.py`, `Diagnostic.py`, `Options.py`, `Windows.py`, `TestWindows.py`
- Added an explicit encoding using this constant to the XML stored in action strings in `Winapp.py` and `TestAction.py`

### DIFF
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -375,17 +375,8 @@ def extended_path(path):
 def free_space(pathname):
     """Return free space in bytes"""
     if 'nt' == os.name:
-        from ctypes import c_ulong, byref, windll
-        if isinstance(pathname, str):
-            pathname = pathname.decode(Common.FSE)
-        freeBytesAvailable = c_ulong()
-        totalNumberOfBytes = c_ulong()
-        totalNumberOfFreeBytes = c_ulong()
-        windll.kernel32.GetDiskFreeSpaceExW(pathname,
-                                            byref(freeBytesAvailable),
-                                            byref(totalNumberOfBytes),
-                                            byref(totalNumberOfFreeBytes))
-        return totalNumberOfFreeBytes.value
+        import psutil
+        return psutil.disk_usage(pathname).free
     mystat = os.statvfs(pathname)
     return mystat.f_bfree * mystat.f_bsize
 
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -41,6 +41,7 @@ import tempfile
 import time
 import ConfigParser
 import Common
+from Common import expanduser, expandvars
 
 if 'nt' == os.name:
     import pywintypes
@@ -354,23 +355,10 @@ def execute_sqlite3(path, cmds):
 def expand_glob_join(pathname1, pathname2):
     """Join pathname1 and pathname1, expand pathname, glob, and return as list"""
     ret = []
-    pathname3 = os.path.expanduser(
-        expandvars(os.path.join(pathname1, pathname2)))
+    pathname3 = expanduser(expandvars(os.path.join(pathname1, pathname2)))
     for pathname4 in glob.iglob(pathname3):
         ret.append(pathname4)
     return ret
-
-
-def expandvars(path):
-    if (2, 5) == sys.version_info[0:2] and 'nt' == os.name:
-        # Python 2.5 should not change $foo when foo is unknown, but
-        # it actually strips it out.
-        import backport
-        expandvars = backport.expandvars
-        return backport.expandvars(path)
-    else:
-        expandvars = os.path.expandvars
-        return os.path.expandvars(path)
 
 
 def extended_path(path):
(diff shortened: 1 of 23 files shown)


marcolagi:
Can't reproduce the failed test on my machine with Windows 7, trying it on a different one with Windows 8.1

az0:
Macro,

Thank you for the patch. From your forked repository I built an installer (with Python 2.5) and tried it on Windows XP. On my regular profile it worked fine, but then I switched to a profile with special characters and got this error:

````
  File "bleachbit\Options.pyo", line 205, in restore
  File "bleachbit\FileUtilities.pyo", line 469, in guess_overwrite_paths
  File "bleachbit\FileUtilities.pyo", line 519, in same_partition
  File "bleachbit\FileUtilities.pyo", line 390, in free_space
error: (123, 'GetDiskFreeSpaceEx', 'The filename, directory name, or volume label syntax is incorrect.')
  File "bleachbit.py", line 46, in <module>
  File "bleachbit\GUI.pyo", line 40, in <module>
  File "bleachbit\Cleaner.pyo", line 37, in <module>
  File "bleachbit\Special.pyo", line 27, in <module>
  File "bleachbit\Options.pyo", line 316, in <module>
  File "bleachbit\Options.pyo", line 68, in __init__
  File "bleachbit\Options.pyo", line 211, in restore
  File "bleachbit\Options.pyo", line 115, in __set_default
  File "bleachbit\Options.pyo", line 244, in set
  File "bleachbit\Options.pyo", line 75, in __flush
  File "bleachbit\General.pyo", line 115, in makedirs
  File "bleachbit\General.pyo", line 115, in makedirs
  File "bleachbit\General.pyo", line 116, in makedirs
WindowsError: [Error 123] The filename, directory name, or volume label syntax is incorrect: 'C:\\Documents and Settings\\g\xe6l?k'
````

In case anyone wants to test, the installers are here
https://sourceforge.net/projects/bleachbit/files/bleachbit/1.13%20beta/macro-e5d8533

Later I will try to test on a newer Windows OS.

**Comment by marcolagi** [comment]
OK... had only tested this on "Marcö", I'll try on "ɡælɪk" as well.

**Comment by marcolagi** [comment]
Summary of changes:
- reimplemented `os.expanduser`, `os.expandvars` and `FileUtilities.expandvars` in `Common.py`
- replace `win32file.GetDiskFreeSpaceEx` with `psutil.disk_usage`
- encode `utf8` or `mbcs` before writing to `ConfigParser`, decode after reading from it
- replace existing environment variables in `os.environ` with their `utf8` encoding
- replace `win32com.client.Dispatch` with `winshell.CreateShortcut`

**Comment by marcolagi** [comment]
The only way I found to get around the limitations of `os.path.expandvars` and `os.path.expanduser` (they're not unicode compatible on windows) is to reimplement them in `Common.py` using `_winreg.ExpandEnvironmentStrings`. This function is only available for python 2.6+, so python 2.5 still uses the old `expandvars` in `backport.py`, which is based on `os.environ`, which is also not unicode compatible. So, paths with non-ascii characters still won't work with python 2.5.

File: bleachbit/FileUtilities.py [hunk 1]
```diff
@@ -375,17 +375,8 @@ def extended_path(path):
...
-            pathname = pathname.decode(Common.FSE)
-        freeBytesAvailable = c_ulong()
-        totalNumberOfBytes = c_ulong()
-        totalNumberOfFreeBytes = c_ulong()
-        windll.kernel32.GetDiskFreeSpaceExW(pathname,
-                                            byref(freeBytesAvailable),
-                                            byref(totalNumberOfBytes),
-                                            byref(totalNumberOfFreeBytes))
-        return totalNumberOfFreeBytes.value
+        import psutil
```
marcolagi: OK, makes sense. The old method is using `win32file.GetDiskFreeSpaceEx`, but it doesn't accept unicode. Then I switched to `windll.kernel32.GetDiskFreeSpaceExW`, but this sometimes gives different free bytes than `wmic`. So, `windll` is not a viable option either, only `win32file` on non-ascii paths would work. The alternative is to always use `wmic` I guess?

az0:
I am testing this and trying to build an installer for others to test too, but as the new code requires an upgrade of the dependencies, I am working through that right now. I have made some progress. Thanks for your patience.

az0:
@marcolagi 

On Windows XP with Python 2.7.12 I launched with the username ɡælɪk, so that is big progress. I was also able to clean Firefox, so that is great.

Then I noticed this error
````
Error in execution of Command to delete C:\Documents and Settings\ɡælɪk\Cookies\index.dat
  File "bleachbit\Worker.pyo", line 84, in execute
  File "bleachbit\Command.pyo", line 89, in execute
  File "bleachbit\Windows.pyo", line 149, in delete_locked_file
error: (123, 'MoveFileEx', 'The filename, directory name, or volume label syntax is incorrect.')
````

Would you please take a look?

marcolagi:
Could not reproduce on Win8, let me find a machine with WinXP.

az0:
To facilitate testing I put binary packages here 

https://sourceforge.net/projects/bleachbit/files/bleachbit/1.13%20beta/marco-be9d2f6%202016-10-16/

az0:
I am closing this pull request.

I paid the contract and merged the changes into a new branch called utf8filenames. It is a separate branch because worked is needed to get Python 2.7 and the new GTK+ version working and optimized for size. 

If you can continue working on any remaining issues (like `delete_locked_file` above), I would appreciate it. On the other hand I understand the contract took longer than you expected.

PR #147, opened 2016-10-05
More flexibility in localizations.xml

tstenner:
Rebased and squashed version of PR#144


tstenner:
One test didn't pass due to a timeout, but the Python 2.6 test passed

PR #149, opened 2016-10-16
Make Unix.py a bit more ready for Python 3

tstenner:
Remove print statements and change "None ==" to " is None" to comply with PEP-8 (and reduce the number of warnings in pycharm)

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -32,7 +32,8 @@ import sys
 
 import FileUtilities
 import General
-from Common import _, autostart_path, logger, launcher_path, RawConfigParser
+import Common
+from Common import _, logger
 
 
 class LocaleCleanerPath:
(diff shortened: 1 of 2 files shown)


bleachbit/Unix.py
```diff
@@ -442,13 +438,11 @@ def apt_autoremove():
 def __is_broken_xdg_desktop_application(config, desktop_pathname):
     """Returns boolean whether application deskop entry file is broken"""
     if not config.has_option('Desktop Entry', 'Exec'):
-        print "info: is_broken_xdg_menu: missing required option 'Exec': '%s'" \
-            % (desktop_pathname)
+        logger.info("is_broken_xdg_menu: missing required option 'Exec': '%s'" % desktop_pathname)
```
az0: Is `logger` defined here?  See example here for how when I converted another module
https://github.com/az0/bleachbit/commit/6f52a2be7ee73aa3def780133298cdb39ee24a5d

az0:
Thank you for working on this. Having code that works well on both Python 2 and 3 is an important step

You probably noticed, but please take a look at
  File "/home/travis/build/az0/bleachbit/tests/TestUnix.py", line 196, in test_start_with_computer

bleachbit/Unix.py
tstenner: It's included from Common.py (line 35), so it should be in scope

tstenner:
The test failure is a bit strange - the functions in Unix.py look in `~/.config/autostart/`, the test looks in `/usr/share/applications`.

tstenner:
Lesson learnt: `from Common import launcher_path` makes a copy of `Common.launcher_path`, which ignores the changes made in the test.

PR #151, opened 2016-10-18
journald cleaner

tstenner:
This also adds a generic binary runner, when I've got time I'll rewrite the yum and apt cleaners to use it.
`FileUtilities.human_to_bytes()` also got a bit more liberal, since journalctl reports freed space as `12.0M`, but with the new test this should be no problem.

### DIFF
--- a/bleachbit/Action.py
+++ b/bleachbit/Action.py
@@ -340,8 +340,8 @@ class Journald(ActionProvider):
     """Action to run 'journalctl --vacuum-time=1'"""
     action_key = 'journald.clean'
 
-    def __init__(self, action_element, action_node):
-        super(Journald, self).__init__(action_node)
+    def __init__(self, action_element):
+        pass
 
     def get_commands(self):
         if FileUtilities.exe_exists('journalctl'):
--- a/bleachbit/Action.py
+++ b/bleachbit/Action.py
@@ -336,6 +336,18 @@ class Ini(FileActionProvider):
             yield Command.Ini(path, self.section, self.parameter)
 
 
+class Journald(ActionProvider):
+    """Action to run 'journalctl --vacuum-time=1'"""
+    action_key = 'journald.clean'
+
+    def __init__(self, action_element, action_node):
+        super(Journald, self).__init__(action_node)
+
+    def get_commands(self):
+        if FileUtilities.exe_exists('journalctl'):
+            yield Command.Function(None, Unix.journald_clean, 'journalctl --vacuum-time=1')
+
+
 class Json(FileActionProvider):
 
     """Action to clean JSON configuration files"""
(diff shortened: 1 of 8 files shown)


az0:
Thank you for the new cleaner and extra thanks for making good reuse of code for apt and yum.

You probably noticed the error on Travis in test_apt_autoclean()

tstenner:
I saw it, but my debian box isn't available for testing today. I assume `apt-get` depends on some more environment variables

tstenner:
The python 2.6 test fails, but since it's only in the Unix part, we can get away with it.

az0:
You wrote that just Python 2.6 is failing, but I see Travis CI with Python 2.6 and 2.7 environments have this error

````
error in handle_cleaner_option() for cleaner id = journald, option XML=<option id="clean">
    <label translate="false">clean</label>
    <description>Clean old system journals</description>
    <action command="journald.clean"/>
  </option>

  File "./bleachbit/CleanerML.py", line 95, in handle_cleaner
    self.handle_cleaner_option(option)
  File "./bleachbit/CleanerML.py", line 143, in handle_cleaner_option
    self.handle_cleaner_option_action(action)
  File "./bleachbit/CleanerML.py", line 173, in handle_cleaner_option_action
    provider = actionplugin(action_node)
TypeError: __init__() takes exactly 3 arguments (2 given)
````
https://travis-ci.org/az0/bleachbit/jobs/169718137#L2056

**Comment by ROCKNROLLKID** [comment]
Seeing the same thing on my side, too.

**Comment by tstenner** [comment]
Sorry, I thought of the previous commit. I'll take a look at it next weekend.

**Comment by az0** [comment]
It's getting closer. 

There is an error here
https://travis-ci.org/az0/bleachbit/jobs/176138775#L1295
```
java.xml validates

journald.xml:29: element action: Schemas validity error : Element 'action', attribute 'command': [facet 'enumeration'] The value 'journald.clean' is not an element of the set {'apt.autoclean', 'apt.autoremove', 'chrome.autofill', 'chrome.databases_db', 'chrome.favicons', 'chrome.history', 'chrome.keywords', 'delete', 'ini', 'json', 'mozilla_url_history', 'office_registrymodifications', 'process', 'sqlite.vacuum', 'truncate', 'win.shell.change.notify', 'winreg', 'yum.clean_all'}.

journald.xml:29: element action: Schemas validity error : Element 'action', attribute 'command': 'journald.clean' is not a valid value of the local atomic type.

journald.xml fails to validate
````

tstenner:
I added `journald.clean` to the CleanerML declaration, but I'm not sure if this is the best way to do it. We might want to support more external commands to clean files and adding a function, subclassing Action and editing the xsd could be a bit unwieldy.

az0:
I see what you mean. Perhaps later you could come up with a PR about the xsd.

PR #153, opened 2016-12-13
Python 3 compatibility

tstenner:
- [X] print statements
- [x]  old-style exceptions
- [X] unicode vs. bytes-fun (other branch)
- [X] misc changes
- [X] actually running it


tstenner:
The Travis build fails with an unrelated error (`configure: error: could not find Python headers`), the AppVeyor build fails at TestMemory because the decorator (`unittest.skipUnless`) can't resolve `os.getuid()` on windows. I'll see what I can do about that.

ROCKNROLLKID:
Are you going to update this to Python 3.6.0 or will it be a few versions behind?

tstenner:
Ubuntu 14.04 LTS is currently on 3.4, so I'll target that.

tstenner:
In theory everything should be ready, but I'll just merge the GTK 3 code and test it.

ROCKNROLLKID:
Very nice work. If you could make a build for me, I would be willing to test on my Windows 10 system. I don't know how to compile stuff from source.

tstenner:
Turns out the huge uproar about the str/bytes-split in Python 3 had a basis. At least I'm already on 86% passing tests on Python 2 with `unicode_literals` turned on

az0:
Please note there is a branch with utf8 fixes that I hope to merge, and I hope there are not merge conflicts wit this PR.

tstenner:
Ok, then I'll wait with the unicode changes until that branch is merged

ROCKNROLLKID:
I am seeing 3 errors in appveyor. All 3 of them are related to the TestWindows.py file. Any chance you could take a look at that?

tstenner:
I'll have to get Windows running on a spare PC, but I've been meaning to do that anyway.

az0:
On the other branch I am improving the Appveyor system to build Windows
binaries from Appveyor. This will help people like yourself who do not have
Windows or who do not have the build environment set up. Currently it seems
to be working well with Python 2.7 and GTK+ 2.

tstenner:
AppVeyor and Travis (at least Python 2.7) run without errors. Do you want to merge this PR first and I'll rebase `utf8filenames` or do you want to merge the other branch first?

az0:
First I'd like to merge utf8filenames to master.

Does this PR allow Python 2 and GTK 2 simultaneously with Python 3 and GTK
3?

tstenner:
GTK 2 and GTK 3 at the same time wouldn't be practical and there's still a lot missing towards Python 3 compatibility, but it's working with Python 2 and GTK 2 at the moment.

ROCKNROLLKID:
Maybe we can release alpha 2 of BleachBit with the utf8 commit and let people test it out, then in a month release alpha 3 with python 3 and gtk 3?

az0:
I am actively working on getting the utf8 branch merged to master. The work in the utf8 branch will ultimately help with Python 3 and GTK 3, as the work includes shrinking the giant Windows Python and GTK+ runtime environments. There are also many changes and benefits there, and I want to make a final release.

After utf8 is merged and an alpha released, we can check whether there are any merged conflicts here, fix those, and check this PR for regressions. I am open to publishing and advertising Python 3 and/or GTK 3 releases of BleachBit, like as alpha releases. Because they need more time for testing and burn in, I will wait before pushing them for a final release.

Thanks!

az0:
I merged the branch, and there are some merge conflicts. Would you please take a look?

bleachbit/Action.py
```diff
@@ -408,15 +409,13 @@ def run_process():
                     rc = 0  # unknown because we don't wait
                     from subprocess import Popen
                     Popen(self.cmd)
-            except Exception, e:
+            except Exception as e:
                 raise RuntimeError(
                     'Exception in external command\nCommand: %s\nError: %s' % (self.cmd, str(e)))
             else:
                 if not 0 == rc:
-                    logger = logging.getLogger(__name__)
```
az0: I remember reading somewhere that defining the logger locally is better than getting it from the module, so I will need to check into this change.

tstenner: With the `getLogger` we can get separate loggers for `bleachbit.foo` and `bleachbit.bar` and collect logs for both with a logger for just `bleachbit`. Currently we don't do this but I'll revert the `getLogger` removals just in case we might want to some day.

az0: OK, thanks.

tests/TestWorker.py
```diff
@@ -249,10 +249,9 @@ def test_InvalidEncoding(self):
         """Test Worker using Action.InvalidEncodingAction"""
         self.action_test_helper('invalid.encoding', 0, 0, 4096, 2, 3, 2)
 
+    @unittest.skipUnless('nt' == os.name, 'skipping on non-Windows')
```
az0: This doesn't look like a Python 3 thing, but the decorator is more elegant. I hadn't noticed that. Thanks.

tests/TestFileUtilities.py
```diff
@@ -740,25 +732,18 @@ def test_wipe_name(self):
...
-
-        try:
-            import sqlite3
-        except ImportError, e:
-            if sys.version_info[0] == 2 and sys.version_info[1] < 5:
-                print "Warning: Skipping test_vacuum_sqlite3() on old Python"
-                return
-            else:
-                raise e
+        import sqlite3
```
az0: This is good. We don't need to make this PR any bigger with more of this, but FYI: I am cleaning out Python 2.5 specific code. I have some more of these clean ups planned after I merged your PR.

az0:
I am wondering how much success you had running BleachBit with Python 3. It looks like there is no such thing as PyGTK 2 for Python 3, and after a quick road test, the CLI has an AttributeError and the tests have an ImportError. I thought I would check with you before further checking

ROCKNROLLKID:
I believe Python and GTK are following the same code paths, so once a new Python major release is out, then GTK will follow, or vise-vera.

tests/TestWorker.py
tstenner: One next step is to include `from __future__ import absolute_import`, but this complicates the imports in tests. Currently, I'm using the [unittest2 test discovery](https://docs.python.org/2/library/unittest.html#test-discovery), but this runs all tests.
Skipping the tests lets me use either the PyCharm test runner or the unittest2 test discovery.

tstenner:
PyGTK (the bindings for GTK 2) aren't compatible with Python 3. It might be possible to use PyGObject with GTK 2 ([but not desirable](http://stackoverflow.com/questions/20110586/may-i-use-latest-pygobject-for-gtk2-with-gobject-introspection-on-windows)), but I don't think it'd be useful.
The GTK 3 changes are necessary for Python 3, but we should take take of the CLI first.

bleachbit/Action.py
tstenner: According to [this SO answer](http://stackoverflow.com/questions/22807972/python-best-practice-in-terms-of-logging) it's best to use a logger instance per module. This way, we only need to change `getLogger(__name__)` to `getLogger('bleachbit.foo')` once per file.

tstenner: On the other hand [this source](https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/) says it's bad. I'll look into it

tstenner: We don't use `logging.config`, so we're not affected by this problem. I also did a quick test with 2 files and file level logger instances and it worked as it should.

az0:
For the merge commit and for the release notes eventually on bleachbit.org do you have a preference for your username or full name?

tstenner:
Doesn't matter to me, whatever fits best with everyone else.

Issue #156, opened 2016-12-21
Mac - Traceback error for below code snippet


brahmastra2016:
i get below error but work fine when i make this below change.

ERROR:

Exception while running operation 'firefox': ''NoneType' object has no attribute 'split''
Traceback (most recent call last):
  File "/Users/nshetty/bleachbit/bleachbit/Worker.py", line 336, in run_operations
    for dummy in self.clean_operation(operation):
  File "/Users/nshetty/bleachbit/bleachbit/Worker.py", line 139, in clean_operation
    if self.really_delete and backends[operation].is_running():
  File "/Users/nshetty/bleachbit/bleachbit/Cleaner.py", line 159, in is_running
    if Unix.is_running(pathname):
  File "/Users/nshetty/bleachbit/bleachbit/Unix.py", line 496, in is_running
    return is_running_darwin(exename)
  File "/Users/nshetty/bleachbit/bleachbit/Unix.py", line 467, in is_running_darwin
    processess = (re.split(r"\s+", p, 10)[10] for p in run_ps().split("\n") if p != "")
AttributeError: 'NoneType' object has no attribute 'split'

CODE:

def is_running_darwin(exename, run_ps=None):
    if run_ps is None:
        def run_ps():
            subprocess.check_output(["ps", "aux", "-c"])
    try:
        processess = (re.split(r"\s+", p, 10)[10] for p in run_ps().split("\n") if p != "")
        next(processess)  # drop the header
        return exename in processess
    except IndexError:
        raise RuntimeError("Unexpected output from ps")

CHANGE:

processess = (re.split(r"\s+", p, 10)[10] for p in  subprocess.check_output(["ps", "aux", "-c"]).split("\n") if p != "")

brahmastra2016:
also work with this below changes

        def run_ps():
            ps = subprocess.check_output(["ps", "aux", "-c"])
            return ps

az0:
If you propose this as a good fix for Darwin, how about a pull request?

az0:
Is this still an issue? If so, please reopen

PR #157, opened 2016-12-24
Traceback error fix - Mac osx [Issue #156]

brahmastra2016:
Traceback error fix for mac osx


az0:
Have you had a chance to look at my code comments? I am just wondering. If you are busy, that is fine. I will wait for your reply.

PR #159, opened 2016-12-24
Delete __init__.py

ROCKNROLLKID:
This file is empty. Anyone know what it was even used for? If not, I think we can just delete this.


az0:
This file is required to make the directory a module

http://stackoverflow.com/questions/10863268/how-is-an-empty-init-py-file-correct

https://travis-ci.org/az0/bleachbit/builds/186470344#L1341

PR #164, opened 2016-12-28
Python 3 compatibility: imports, types (WIP)

tstenner:
This changes the imports to absolute imports and reorders them according to PEP8 and changes tests to use Python 3 compatible types (e.g. `six.string_types` instead of `(str,unicode)`).


az0:
Today I committed a change that make the unit tests on the master branch pass again, so all pull requests won't have the disadvantage of always failing on Appveyor (issue #165). This caused merge conflict here---probably a minor one. Otherwise, I will try to avoid sweeping changes in the near future that would cause merge conflicts with this.

I am looking forward to these changes, but I know the string handling can be difficult. Please let me know when this pull request progresses from work in progress to complete. Thank you!

bleachbit/Action.py
```diff
@@ -165,8 +165,7 @@ def get_file(path):
 
         def get_walk_all(top):
             for expanded in glob.iglob(top):
-                for path in FileUtilities.children_in_directory(
-                        expanded, True):
+                for path in FileUtilities.children_in_directory(expanded, True):
```
az0: This looks like just a line break change? Usually I use `autopep` with the defaults. It's in the `makefile`.

tests/TestAll.py
```diff
@@ -23,64 +23,15 @@
...
-def suite():
-    """Combine all the suites into one large suite"""
-    suite_ = unittest.TestSuite()
-    map(suite_.addTest, suites)
-    return suite_
+import sys
 
 if __name__ == '__main__':
-    success = unittest.TextTestRunner(verbosity=2).run(suite()).wasSuccessful()
+    print("""You should use the unittest discovery, it's much nicer:
```
az0: OK

az0:
I can see a lot of hard work here and good changes. You integrated the six module, improved unit tests, fixed resource leaks, etc. Thank you. 

You merged in the GTK 3 changes that removes GTK 2 support. In the master branch I would like to support GTK 2 a bit longer. I propose to either support GTK 2 and GTK 3 in the master branch with separate files (like GUI2.py and GUI3.py) or to merge all this into a new branch. What do you think?

tstenner:
The problem with the GTK 2 code is that I can't test Python 3 code with GTK 2 and some code that shoudn't require GTK imports it indirectly. This still needs a lot of work and testing, so combining it with the needed change to GTK 3 should be easier than doing both separately.

I don't think supporting both GTK versions at once is a good idea, as we'll need to drop GTK2 support some day and switch to something else. Also there aren't that many unit tests for the GUI, and most developers will just test their changes against one GUI.

ROCKNROLLKID:
@az0 Maybe we can merge this along with GTK 3 with the modernization branch so people can start testing/building/etc?

az0:
I know it's been too long, sorry. BleachBit 2.0 is 99.999% done, and I
would like to give this the attention it deserves after BleachBit 2.0 is
out.

az0:
The gtk3 branch is the focus of development going forward. As suggested here, it is hard break because it drops all support for GTK 2. I'd like to release maybe one stable BleachBit version with GTK2 and Python 2 and then do a another hard break by dropping Python 2 support.

az0:
Sorry, it took me too long to get to this, and this PR has gotten stale. The good nes is  there will be code commited to update to Python 3, while dropping all support for Python 2.

PR #173, opened 2017-01-14
VLC cleanup

ROCKNROLLKID:
- Removed unneeded white spaces.
- Removed website as it doesn't belong here really.
- Moved contributor to top so it matches with other cleaner files.


ROCKNROLLKID:
By the way, how do I do do several commits to multiple files in 1 PR?

az0:
To do several commits to multiple files, put all the commits in the same branch on your side. For example, this PR is coming from your patch-5 branch

Alternatively alter multiple files in one commit, if the commits are logically related as a single unit

PR #174, opened 2017-01-14
Octave cleanup

ROCKNROLLKID:
- Removed comments. Ubuntu version 12 will be obsolete in BleachBit, soon and version 3.x is 5 years old.


az0:
Thank you, but I would rather leave the old comment: see #175.

PR #175, opened 2017-01-14
Sqlite cleanup

ROCKNROLLKID:
- Removed comment. Too old of version for testing and Ubuntu version 12 will be obsolete in BleachBit, soon.


az0:
Rather than remove it, I updated the comment. Even though BleachBit may not work on old operating systems, it is generally preferable to clean old files that may have been migrated from those systems via upgrade. It is good to know the file this cleaner cleans is stable and well tested.

PR #178, opened 2017-01-21
Absolute imports for Py3

tstenner:
This changes the imports in unit tests to absolute imports


az0:
Nice. It looks like you are really close.

tstenner:
Well, for now it's just tiny parts and the unicode/bytes split is still causing trouble so I try to get the easy stuff merged first so I don't have as many rebase conflicts

PR #184, opened 2017-01-23
Absolute imports part 2

tstenner:
This is the second part of the import changes.


tstenner:
`py2exe` still has a problem with absolute imports. I won't have access to a Windows PC before the weekend, so if anyone could take a look at it that'd be great.

tstenner:
Apparently there are some circular references in the imports that cause problems with relative imports (`from bleachbit import Common`). Is it ok if I move the code in `Common.py` to `__init__.py` and `FileUtilities.py` to resolve this?

az0:
You need to move all the code out of Common.py and delete Common.py?

tstenner:
In case of circular references (`a`: `from b import foo`, `b`: `from a import bar`) the interpreter throws an error, but absolute references (`import b.foo`) are fine. Since `bleachbit.Common.foo` is quite long, this would abbreviate it to `bleachbit.foo`. Also, some functions (`expanduser`, `expandvars`) are better suited for `FileUtilities.py`

az0:
OK, that makes sense. At a glance these changes look good. When AppVeyor CI is passing and you think this is ready, I will take a closer look. I value this progress. Thank you.

bleachbit/Common.py
```diff
@@ -1,344 +0,0 @@
-# vim: ts=4:sw=4:expandtab
```
tstenner: This is now in `__init__.py`

bleachbit/Action.py
```diff
@@ -23,23 +23,21 @@
...
-import General
-import Special
-import Common
 
-from Common import _, FSE, expanduser, expandvars
 
 if 'posix' == os.name:
     re_flags = 0
-    import Unix
+    from bleachbit import Unix
```
tstenner: When there are no circular imports we can translate imports like this

bleachbit/GUI.py
```diff
@@ -428,8 +426,9 @@ def about(self, __event):
         dialog.set_comments(_("Program to clean unnecessary files"))
         dialog.set_copyright("Copyright (C) 2008-2017 Andrew Ziem")
         try:
-            dialog.set_license(open(license_filename).read())
-        except:
+            with open(bleachbit.license_filename) as f:
```
tstenner: This is in the wrong PR, but still a good idea. Otherwise we leak a file descriptor every time a user opens the about dialog

bleachbit/Windows.py
```diff
@@ -202,11 +200,11 @@ def delete_registry_key(parent_key, really_delete):
 
 def delete_updates():
     """Returns commands for deleting Windows Updates files"""
-    windir = expandvars('$windir')
+    windir = bleachbit.expandvars('$windir')
```
tstenner: We can't use `from bleachbit import expandvars` because that'd cause a circular import

tstenner:
This should be everything. I have left comments at the non-obvious changes and fixes for two leaks I accidentally included.

bleachbit/GUI.py
ROCKNROLLKID: Good catch. One of BleachBit's goal is to leave a very minimal footprint, so this is a nice find.

PR #187, opened 2017-01-30
Reduce Windows installation size by ~12M

mkhon:
- Switch to UPX 3.93: this allows to compress all *.pyd and *.dll
(see https://github.com/upx/upx/issues/42).
This saves about 12M of uncompressed distribution size.
- Remove the following unused files:
	- etc\bash_completion.d
	- lib\GNU.Gettext.dll (.Net library)
	- lib\gtk2.0\include


ROCKNROLLKID:
We are about to make a change to gtk/python 3, so not sure if doing this for gtk/python 2 is really worth doing. There has been some discussion about making 2 different branches for gtk/python 2/3, but nothing has been confirmed, yet.

az0:
ROCKNROLLKID, I requested this, and at a glance it seems reusable for Python 3 / GTK 3.

@mkhon: I will take a look. Thank you.

ROCKNROLLKID:
My apologies. Continue on.

PR #191, opened 2017-02-05
Improve unit tests

tstenner:
For an eventual py3-transition we need better unit test coverage.

What's planned / done:
- [X] Introduce a custom base class with better tempfile handling and asserts
- [ ] Convert asserts / tempfile creation to use the class
- [x] Convert asserts from `self.assert_` to more meaningful asserts (especially `self.assert_(isinstance(...))`
- [ ] test for unicode / bytes explicitly

### DIFF
--- a/tests/TestAction.py
+++ b/tests/TestAction.py
@@ -90,7 +90,7 @@ def dir_is_empty(dirname):
     return not os.listdir(dirname)
 
 
-class ActionTestCase(unittest.TestCase, common.AssertFile):
+class ActionTestCase(common.BleachbitTestCase):
 
     """Test cases for Action"""
 
--- a/tests/TestAction.py
+++ b/tests/TestAction.py
@@ -107,8 +107,7 @@ class ActionTestCase(common.BleachbitTestCase):
                 provider = actionplugin(action_node)
         self.assertNotEqual(provider, None)
         for cmd in provider.get_commands():
-            self.assert_(
-                isinstance(cmd, (Command.Delete, Command.Ini, Command.Json, Command.Function)))
+            self.assertIsInstance(cmd, (Command.Delete, Command.Ini, Command.Json, Command.Function))
             if 'process' != command:
                 # process does not have a filename
                 self.assertLExists(filename)
@@ -131,8 +130,7 @@ class ActionTestCase(common.BleachbitTestCase):
             else:
                 raise RuntimeError("Unknown command '%s'" % command)
         if 'walk.all' == search:
-            self.assert_(dir_is_empty(filename),
-                         'directory not empty after walk.all: %s' % filename)
+            self.assertTrue(dir_is_empty(filename), 'directory not empty after walk.all: %s' % filename)
 
     def test_delete(self):
         """Unit test for class Delete"""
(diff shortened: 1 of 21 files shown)


tests/common.py
```diff
@@ -26,24 +26,51 @@
...
 
+import unittest
+import shutil
 import tempfile
 import os
+import os.path
 
 
-class AssertFile:
+class BleachbitTestCase(unittest.TestCase):
```
az0: It's a good idea to use inheritance here

tests/TestAction.py
```diff
@@ -209,10 +202,7 @@ def execute_json(path, address):
 
     def test_process(self):
         """Unit test for process action"""
-        if 'nt' == os.name:
-            cmd = 'cmd.exe /c dir'
-        if 'posix' == os.name:
-            cmd = 'dir'
+        cmds = {'nt': 'cmd.exe /c dir', 'posix': 'dir'}
```
az0: This is a creative use of a dictionary. Nice.

tests/TestCommand.py
```diff
@@ -21,70 +21,52 @@
...
-        self.assert_(os.path.getsize(path) > 0)
+        self.assertExists(path)
+        self.assertGreater(os.path.getsize(path), 0)
 
         # preview
         ret = cmd.execute(False).next()
-        self.assert_(os.path.exists(path))
-        self.assert_(os.path.getsize(path) > 0)
+        self.assertExists(path)
+        self.assertGreater(os.path.getsize(path), 0)
```
az0: assertGreater is another benefit of Python 2.7+

tests/TestFileUtilities.py
```diff
@@ -419,42 +392,34 @@ def test_expand_glob_join(self):
...
         else:
             # unchanged
             tests = (('/home/foo', '/home/foo'),)
-            tests_undo = tests
-        for test in tests:
-            self.assertEqual(extended_path(test[0]), test[1])
-        for test in tests_undo:
-            self.assertEqual(extended_path_undo(test[0]), test[1])
+        for short, extended in tests:
+            # already extended path shouldn't be changed
```
az0: These are logical tests

az0:
This is great work. I love a cleanup like this that reduces code. 

I could merge this now, but would you rather wait? I see only 2/4 items are checked off.

ROCKNROLLKID:
I am seeing merge conflicts with this. Will need to fix that first.

tstenner:
I'd rather go through the last remaining files first and then rebase the PR, after that I'd also be in favor of merging it (less conflicts for me and everyone else). I can always open another PR for the remaining items.

ROCKNROLLKID:
Are we holding back Python 3 coding for the next update then?

az0:
@tstenner  OK, please write a note when it's ready.

@ROCKNROLLKID I would like to hold major changes including Python 3 and GTK 3 until after BleachBit 2.0 is released. Then it may also take a follow-up maintenance release to fix important regressions. After that I would like to evaluate how well Python 3 and GTK 3 work, especially on Windows. Given the recent work to shrink the new GTK 2 on Windows for BleachBit 2.0, I assume that I can also shrink Python 3 and GTK 3. This pull request is a stepping stone for Python 3 because it improves test coverage.

tstenner:
The test for locked files fails because the worker reports that it cleaned too many bytes (6 instead of 3), but I (think I) didn't change anything relevant in the failing test. Could someone with a windows installation take a look at it?

az0:
The error 6 != 3 looks like something I broke related to the new WindowsWipe code. I will work on it.

az0:
The 6 != 3  error was fixed in commit 193e07a

tstenner:
The unit tests pass now, travis hits a time limit somehow (downloading the appveyor artifacts?). We should merge this, I'll tackle the rest of the todo-list in a separate PR

az0:
Yes, Travis hit a time limit waiting on AppVeyor, while Appveyor failed because of the errno typo.

PR #192, opened 2017-02-07
Fix bug in network share path handling

tstenner:
UNC paths have only one backslash before the server name.

> The "\\?\" prefix can also be used with paths constructed according to the universal naming convention (UNC). To specify such a path using UNC, use the "\\?\UNC\" prefix. For example, "\\?\UNC\server\share", where "server" is the name of the computer and "share" is the name of the shared folder. 
[source](https://msdn.microsoft.com/en-us/library/aa365247(VS.85).aspx#maxpath)


tstenner:
The Travis CI failed because of a network error. Strange.

PR #213, opened 2017-04-13
Close running applications

brahmastra2016:
Added User select able option to close running application.

Default= “NO”

### DIFF
--- a/bleachbit/Cleaner.py
+++ b/bleachbit/Cleaner.py
@@ -27,7 +27,7 @@ from __future__ import absolute_import, print_function
 from bleachbit import _, expanduser, expandvars
 from bleachbit.FileUtilities import children_in_directory
 from bleachbit.Options import options
-from bleachbit import Command, FileUtilities, Memory, Special
+from bleachbit import Command, FileUtilities, Memory, Special,GuiBasic
 
 import glob
 import logging
@@ -35,6 +35,8 @@ import os.path
 import re
 import sys
 import warnings
+import subprocess
+
 
 if 'posix' == os.name:
     from bleachbit import Unix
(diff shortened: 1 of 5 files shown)


ROCKNROLLKID:
This might be nice to have for Windows, too, especially when cleaning browsers. Looks good otherwise, thought I don't own a Mac to actually test this.

az0:
@ROCKNROLLKID 
The title says "Mac" but the code has paths for all operating systems. 

@brahmastra2016 
Thank you for working on this! It does seem useful. 

The major things I see now are
1.  Checking the command line arguments for `--preset` is not a reliable way to detect command line mode. For example, command line mode can be run with just `bleachbit --preview google_chrome.cache`. 

The heuristic I see is: if there are zero command line arguments, it is GUI mode. If the command line has --gui, it is GUI. Otherwise it is CLI. 

Rather than code this rule many places, there should be a standard way. One option is to create a new function, like in `bleachbit/__init__..py`. Another option is setting a variable in that module.

2. There are two kinds of command line mode. If there is no terminal, such as when the application is invoked from cron, then the application should not prompt with a question because there cannot be a response. I think `sys.stdout.isatty()` is the right way to check.

My lesser-important feedback: There are many nested conditions and much similarity between operating systems, so refactoring would make this look cleaner and shorter.

brahmastra2016:
I haved added it for windows too , just forgot to mention it.

ROCKNROLLKID:
I feel like closing ALL running applications might be too aggressive. What if we just limited this to web browsers? That seems like it is the only time we really need to close any applications is when we are cleaning web browsers.

brahmastra2016:
That is a good suggestion,  i am planning to put some time into this project.been off for a while now.
--
Sent from myMail for Android Friday, 15 December 2017, 05:33PM -05:00 from Alex  notifications@github.com :

ROCKNROLLKID:
Another theory I had is to do what CCleaner does and just notify the user that a browser is running and ask if BleachBit should terminate the process. This is probably the best way to do this.

brahmastra2016:
That is what my current code does it prompt user if they wish to close the app or continue  without closing.
--
Sent from myMail for Android Wednesday, 20 December 2017, 09:12PM -05:00 from Alex  notifications@github.com :

az0:
This is an interesting PR, so I hope you can please make a few adjustments in response to my comment https://github.com/bleachbit/bleachbit/pull/213#issuecomment-293766999 dated April 12. Until then, I will close it.

PR #223, opened 2017-07-11
Drop RPM dependency on gnomevfs

Mailaender:
According to 1c4a117ef2608a53001e5ecf9c9a3ce2653c38e9 and 89a7a06f6f4bfe734d3d1933058273e337669d9d the code was migrated. See also https://build.opensuse.org/request/show/509524


az0:
@Mailaender  Thank you for the patch and the well-written description.

PR #252, opened 2017-11-12
Update translation pt_BR

thezilion:
Update translations in pt_BR


az0:
Thank you for working on this translation. Would you please make translation changes at https://translations.launchpad.net/bleachbit/trunk/+pots/bleachbit/pt_BR/+translate ? This is where I am managing translations, and changes flow in a single direction only, so if I merge your translation now, it would get overwritten by the next change there.

thezilion:
Of Course :)

Thanks!

PR #347, opened 2018-05-22
If user selects a single cleaner option, display action summary in right pane

fxtentacle:
Currently, there's no difference between selecting a group or selecting a single option in the left pane. These two patches change the behavior so that if a single option is selected in the left pane, then a textual description of the associated actions for that option is displayed in the right pane. This allows power users to quickly review the files or paths that a given rule will modify and is more concise than the output produced by preview.

### DIFF
--- a/bleachbit/Action.py
+++ b/bleachbit/Action.py
@@ -104,7 +104,8 @@ class ActionProvider:
         """Yield each command (which can be previewed or executed)"""
         pass
 
-
+    def __str__(self):
+        return self.action_key
 #
 # base class
 #
@@ -194,6 +195,21 @@ class FileActionProvider(ActionProvider):
         import itertools
         for f in itertools.ifilter(self.path_filter, self._get_paths()):
             yield f
+
+    def __str__(self):
+        ret = self.action_key+":   "
+        ret += "and   ".join(self.paths)
+        if self.regex:
+            ret += "   if   "+self.regex
+        if self.nregex:
+            ret += "   if not   "+self.nregex
+        if self.wholeregex:
+            ret += "   if whole   "+self.wholeregex
+        if self.nwholeregex:
+            ret += "   if not whole   "+self.nwholeregex
+        if self.object_type:
+            ret += "   if type   "+self.object_type
+        return ret
 
     def _get_paths(self):
         """Return a filtered list of files"""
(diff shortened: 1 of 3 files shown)


bleachbit/Cleaner.py
```diff
@@ -134,6 +134,19 @@ def get_option_descriptions(self):
         if self.options:
             for key in sorted(self.options.keys()):
                 yield (self.options[key][0], self.options[key][1])
+                
+    def get_long_option_description(self, key):
+        """Yield the name and a potentially more verbose description for a single option"""
+        if self.options:
+            ret = self.options[key][1]
+            ret += "\n\nThis cleanup option will perform the following actions:\n\n"
```
az0: This string should be eligible for localization/translation. Wrap it in the `_()` function.

bleachbit/Action.py
```diff
@@ -195,6 +196,21 @@ def get_paths(self):
         for f in itertools.ifilter(self.path_filter, self._get_paths()):
             yield f
 
+    def __str__(self):
+        ret = self.action_key+":   "
+        ret += "and   ".join(self.paths)
```
az0: Under Google Chrome I see

`delete:   /home/z/.config/chromium/Safe Browsing Channel IDs-journaland   %LocalAppData%\Chromium\User Data/Safe Browsing Channel IDs-journal`

There is missing whitespace before "and"

bleachbit/Cleaner.py
```diff
@@ -134,6 +134,19 @@ def get_option_descriptions(self):
         if self.options:
             for key in sorted(self.options.keys()):
                 yield (self.options[key][0], self.options[key][1])
+                
+    def get_long_option_description(self, key):
+        """Yield the name and a potentially more verbose description for a single option"""
```
az0: Under the System category there are no actions listed. Either they should be listed, or the introductory phrase "this cleanup option" should be omitted.

Likewise for the Firefox cleaner. (I have recently done prep work to move it CleanerML, which would fix this, but that is not quite ready.)

az0:
@fxtentacle  Thanks for the interesting PR. I am adding review comments. Please take a look. Also, Codacy noticed some issues: these are not critical, but it would be nice if you ran `autopep8` on the lines you changed.

az0:
Please add a more detailed description for the Process action provider used for Windows thumbnails.

az0:
@fxtentacle Have you had a chance to look at my comments? I am hoping to push out a beta release in the not too distant future

az0:
This looks like an itneresting pull request. Please reopen after responding to the concerns.

PR #370, opened 2018-08-06
Add OpenBSD cleaner path

### LINKED ISSUE #371: Updates to run on OpenBSD?
Looks like the `__init__.py` is missing a directive, though one exists for NetBSD. I did PR #370  to help this.

In `FileUtilities.py` there are explicit checks for Linux or Mac systems. Will either of these be viable for OpenBSD?

I do not mind testing possible resolutions to this.

evitalis:
Add a path for cleaners on OpenBSD.
Was not sure what else may be needed to be changed but this at least allows it to be installed for further testing.

Partially resolves #371


az0:
You mentioned "allows for further testing," but please keep in mind that BleachBit should be easy to run from source without installation. In any case, thanks for the PR.

PR #376, opened 2018-08-08
Fix Windows Explorer search history cleaner on Windows 10

mgziminsky:
At least on Windows 10, the "Search History" cleaner under "Windows Explorer" doesn't clear the file explorer search history. To do so, it just needs to delete an additional registry key.


az0:
Closes https://bugs.launchpad.net/bleachbit/+bug/1261098

Thank you, @mgziminsky !

PR #393, opened 2018-10-02
Gnome notifications fix

Djaler:
Due to Gnome Guidelines https://wiki.gnome.org/Initiatives/GnomeGoals/NotificationSource

Now, due to the absence of these parameters, notifications from the bleachbit are displayed as "Others"


az0:
@djaler thanks! Have you had a chance to look at the CI error?

az0:
Ah, `WindowsError: [Error 123] The filename, directory name, or volume label syntax is incorrect` is a random error. It can be ignored.

PR #403, opened 2018-10-31
Improve cleaner support for FreeBSD

luoliyan:
The `os` option in CleanerML works quite well, but it disabled quite a few of the cleaners that work on FreeBSD. I've re-enabled the cleaners that I've tested on FreeBSD, and added some missing paths.

The results of running the test suite are unchanged: `failures=12, errors=3, skipped=35`

### DIFF
--- a/bleachbit/CleanerML.py
+++ b/bleachbit/CleanerML.py
@@ -92,6 +92,8 @@ class CleanerML:
             current_os = ('bsd', 'openbsd', 'unix')
         elif platform.startswith('netbsd'):
             current_os = ('bsd', 'netbsd', 'unix')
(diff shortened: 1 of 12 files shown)


az0:
@luoliyan Looks great except the .XSD needs to be updated for this error

`libreoffice.xml:26: element running: Schemas validity error : Element 'running', attribute 'os': [facet 'enumeration'] The value 'freebsd' is not an element of the set {'bsd', 'darwin', 'linux', 'netbsd', 'openbsd', 'unix', 'windows'}.`

luoliyan:
Well that's was an interesting oversight.

PR #407, opened 2018-11-11
Install for all users

teknowledgist:
This should make the install a "proper" system install rather than a user install for administrators.

### DIFF
--- a/windows/bleachbit.nsi
+++ b/windows/bleachbit.nsi
@@ -204,8 +204,8 @@ SectionEnd
 
 
 SectionGroup /e Shortcuts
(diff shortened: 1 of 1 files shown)


az0:
Thanks for the patch. Please check the error log

2018-11-11 03:59:54 - ERROR - Error: command SetShellVarContext not valid outside Section or Function
Error in script "windows\bleachbit.nsi" on line

az0:
Thanks for the PR and for the encouragement to address this issue. I took a different approach here 2c05f3feb6d7a7213ee3805767b81235f4726e7b

PR #427, opened 2018-12-20
vuze.xml

Tobias-B-Besemer:


cleaners/vuze.xml
```diff
@@ -30,7 +33,7 @@
   </option>
   <option id="cache">
     <label>Cache</label>
-    <description>Delete the cache</description>
+    <description>Delete the IPFilter Cache</description>
```
az0: The old string is already translated into dozens of  languages. The distinction about IPFilter is not necessary because Vuze has only one kind of cache, and it's also shown in the filename in the preview

az0:
Please see the earlier comment

PR #453, opened 2019-02-26
Deepscan

Tobias-B-Besemer:


az0:
Deleting *.log can be dangerous, as you noted in the warning. However, I'm afraid too many people won't read this warning, so I suggest to move this to the CleanerML repository for advanced users. Feel free to create a deepscan_extra.xml or deepscan_advanced.xml there.

By the way
1.  Only warning XML element is supported.
2. I'd rather not change the existing descriptions because the description doesn't add much new information that isn't in the label (like about Thumbs.db). The new strings would require 120 translations (2 strings * about 60 languages)

PR #462, opened 2019-03-06
Vuze

Tobias-B-Besemer:
Seems I forgot to make a pull request... :-/

### DIFF
--- a/cleaners/vuze.xml
+++ b/cleaners/vuze.xml
@@ -32,14 +32,14 @@
     <action command="delete" search="glob" path="$APPDATA\Azureus\active\*.bak"/>
   </option>
   <option id="cache">
-    <label>IPFilter Cache</label>
+    <label>Cache</label>
     <description>Delete the IPFilter Cache</description>
     <action command="delete" search="file" path="~/.azureus/ipfilter.cache"/>
     <action command="delete" search="file" path="$APPDATA\Azureus\ipfilter.cache"/>
   </option>
   <option id="logs">
     <label>Logs</label>
-    <description>Delete all logs (inlc. tracker and update)</description>
+    <description>Delete the logs</description>
     <action command="delete" search="walk.all" path="~/.azureus/logs/"/>
     <action command="delete" search="file" path="~/.azureus/update.log"/>
     <action command="delete" search="file" path="~/.azureus/tracker.log"/>
--- a/cleaners/vuze.xml
+++ b/cleaners/vuze.xml
@@ -19,11 +19,11 @@
     along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
     @app Vuze
-    @url http://www.vuze.com/
+    @url https://www.vuze.com/
     @os Windows, Linux
     @cleanerversion v2.0.0
-    @cleanerdate 2019-03-18.2
-    @cleanerby Andrew Ziem (2009-02-15 - 2009-10-04) & Tobias B. Besemer (2018-12-20 - 2019-03-18)
+    @cleanerdate 2019-03-19.2
+    @cleanerby Andrew Ziem (2009-02-15 - 2009-10-04) & Tobias B. Besemer (2018-12-20 - 2019-03-19)
     @tested ok v???, Windows 7
     @testeddate ???
     @testedby https://github.com/Tobias-B-Besemer
@@ -41,13 +41,22 @@
     <!--
     Windows itself changes %ProgramFiles% for 32-bit processes running in a 64-bit operating system as documented here:
     https://docs.microsoft.com/en-us/windows/desktop/winprog64/wow64-implementation-details
-    ...so we use %SystemDrive%\Program Files\ & %SystemDrive%\Program Files (x86)\...
-    -->
-    <value os="windows">%SystemDrive%\Program Files\Vuze</value>
-    <value os="windows">%SystemDrive%\Program Files (x86)\Vuze</value>
+    ...so we used "%SystemDrive%\Program Files\" & "%SystemDrive%\Program Files (x86)\"...
+    <value os="windows">%SystemDrive%\Program Files\[ProgramName]</value>
+    <value os="windows">%SystemDrive%\Program Files (x86)\[ProgramName]</value>
+    ...but %ProgramFiles% & %ProgramFiles(x86)% don't need to be on %SystemDrive% !!!
+    So we use: -->
+    <value os="windows">%ProgramW6432%\Vuze</value>
+    <!-- ...that BleachBit as 32-bit program can find "\Program Files\Vuze" and... -->
+    <value os="windows">%ProgramFiles(x86)%\Vuze</value>
+    <!-- ...that BleachBit can find "\Program Files (x86)\Vuze" and... -->
+    <value os="windows">%ProgramFiles%\Vuze</value>
+    <!-- ...that BleachBit can find "\Program Files\Vuze" on a x86 system...
+    ...and later, as 64-bit program, "\Program Files\Vuze" (x64)!
+    So all cases solved! ;-) Tobias. -->
     <!--
     Linux $$ProgramFiles$$ not yet figured out! I will do this later! Tobias.
-    <value os="linux">~/.azureus</value>
+    <value os="linux">~/.vuze</value>
     -->
   </var>
   <option id="backup">
--- a/cleaners/vuze.xml
+++ b/cleaners/vuze.xml
@@ -18,60 +18,70 @@
     You should have received a copy of the GNU General Public License
     along with this program. If not, see <http://www.gnu.org/licenses/>.
 
-	Rules in CleanerML written by https://github.com/Tobias-B-Besemer
-	v0.5.5 done on 2019-02-26
+    @url http://www.vuze.com/
+    @cleanerversion v2.0.0
+    @cleanerdate 2019-03-11
+    @cleanerby https://github.com/Tobias-B-Besemer
+    @tested ok 
+    @testeddate 
+    @testedby 
+    @note 
 
 -->
 <cleaner id="vuze">
   <label>Vuze/Azureus</label>
   <description>File transfer client</description>
-  <option id="backup_files">
+  <var name="profile">
+    <value search="glob" os="windows">$APPDATA\Azureus</value>
+    <value search="glob" os="linux">~/.azureus</value>
+  </var>
+  <var name="programfiles">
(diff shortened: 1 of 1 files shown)


az0:
Thank you, Tobias. To not delay BleachBit 2.2 with needs of more testing, I marked this for release 2.4

Tobias-B-Besemer:
Cleaner extremely slows down start process of BleachBit (log stays for some sec. after "vim"), and I don't know why, yet.

cleaners/vuze.xml
```diff
@@ -52,25 +55,26 @@
...
-    <action command="delete" search="walk.all" path="$$profile$$\stats\"/>
-    <action command="delete" search="glob" path="$$profile$$\stats\*"/>
     <action command="delete" search="file" path="$$profile$$\update.log"/>
     <action command="delete" search="file" path="$$profile$$\tracker.log"/>
     <action command="delete" search="glob" path="$$programfiles$$\*.log"/>
     <action command="delete" search="walk.files" path="$$programfiles$$\log\"/>
     <action command="delete" search="glob" path="$$programfiles$$\.install4j\*.log"/>
+    <!-- Delete the stats, too: -->
+    <action command="delete" search="walk.all" path="$$profile$$\stats\"/>
+    <action command="delete" search="glob" path="$$profile$$\stats\*"/>
```
az0: This is a similar scenario to other pull reqeusts. Line 77 won't match anything because line 76 deleted the contents already. If you want to delete the folder itself, change line 77 from `glob` to `file` and remove the wildcard from `path`

cleaners/vuze.xml
```diff
@@ -82,8 +86,8 @@
     <action command="delete" search="glob" path="$$profile$$\torrents\*.tmp._az"/>
     <action command="delete" search="glob" path="$$profile$$\torrents\*.torrent._az.saving"/>
     <action command="delete" search="walk.all" path="$$profile$$\updates\"/>
-    <action command="delete" search="walk.all" path="%LocalAppData%\Temp\AzureusInstall\"/>
-    <action command="delete" search="glob" path="%LocalAppData%\Temp\AzureusInstall\"/>
-    <action command="delete" search="glob" path="%LocalAppData%\Temp\core_Azureus*"/>
+    <action command="delete" search="walk.all" path="%TMP%\AzureusInstall\"/>
+    <action command="delete" search="glob" path="%TMP%\AzureusInstall\"/>
+    <action command="delete" search="glob" path="%TMP%\core_Azureus*"/>
```
az0: The environment variable change looks fine. On line 90 search=glob should be `search="file"` because there is no wildcard in `path`

cleaners/vuze.xml
```diff
@@ -18,14 +18,16 @@
     You should have received a copy of the GNU General Public License
     along with this program. If not, see <http://www.gnu.org/licenses/>.
 
+    @app Vuze
     @url http://www.vuze.com/
+    @os Windows, Linux
     @cleanerversion v2.0.0
-    @cleanerdate 2019-03-11
+    @cleanerdate 2019-03-13
     @cleanerby https://github.com/Tobias-B-Besemer
```
az0: I created this in 2009, but the credit just mentions you. 

😢

https://github.com/bleachbit/bleachbit/commit/efb1b7de9409a59bae7dba40db6f57f0557ef545#diff-e3a08ce16efae5e4e6cbfcef163ffe87

cleaners/vuze.xml
Tobias-B-Besemer: I will try to fix that... (...in all my cleaners!)
...but I have filled an issue about such problems for me here: https://github.com/bleachbit/wishlist/issues/33
Would be nice to get there a feedback! Thx in advance!

cleaners/vuze.xml
Tobias-B-Besemer: Sorry, I c&p the header! Seems I just forgot to add "??? & " before (like I did on other cleaners)! I add always "??? & " because I don't know who already worked on them...
I will add you! Someone else, too?

Tobias-B-Besemer: Is there any chance to browse the history of a single file on git to see who have made PR to this file ???

Tobias-B-Besemer: I guess I found it... https://github.com/bleachbit/bleachbit/commits/master/cleaners/vuze.xml

cleaners/vuze.xml
```diff
@@ -34,11 +34,16 @@
...
     <value search="glob" os="linux">~/.azureus</value>
   </var>
   <var name="programfiles">
-    <value search="glob" os="windows">$ProgramFiles\Vuze</value>
+    <!--
+    Windows itself changes %ProgramFiles% for 32-bit processes running in a 64-bit operating system as documented here:
+    https://docs.microsoft.com/en-us/windows/desktop/winprog64/wow64-implementation-details
+    ...so we use %SystemDrive%\Program Files\...
+    -->
+    <value search="glob" os="windows">%SystemDrive%\Program Files\Vuze</value>
```
az0: `ProgramW6432` is better than `%SystemDrive%\Program Files`

cleaners/vuze.xml
az0: Yes, the Git history is helpful. Caveat: Older change records were imported from SVN when the authors were written in the commit body instead of the commit header.

cleaners/vuze.xml
Tobias-B-Besemer: But what's about W32 Support ???

az0: I don't understand your question. `ProgramW6432` points to `c:\program files\` regardless of whether BleachBit is 32-bit or 64-bit.

Tobias-B-Besemer: I talk about the install path of Vuze on 32-bit systems... AFAIK is ProgramW6432 not available in Windows 32-bit...

cleaners/vuze.xml
Tobias-B-Besemer: I changed that!

cleaners/vuze.xml
Tobias-B-Besemer: I miss the old COBOL times when programmers got payed per line the written... ;-) :-D

Tobias-B-Besemer:
Damn! Something was going wrong!

cleaners/vuze.xml
```diff
@@ -18,34 +18,91 @@
...
+    <value os="linux">~/.azureus</value>
+  </var>
+  <var name="ProgramFiles">
+    <!--
+    Windows itself changes %ProgramFiles% for 32-bit processes running in a 64-bit operating system as documented here:
+    https://docs.microsoft.com/en-us/windows/desktop/winprog64/wow64-implementation-details
+    ...so we use %SystemDrive%\Program Files\ & %SystemDrive%\Program Files (x86)\...
+    -->
+    <value os="windows">%SystemDrive%\Program Files\Vuze</value>
+    <value os="windows">%SystemDrive%\Program Files (x86)\Vuze</value>
```
az0: According to this [Wikipedia article on Program Files](https://en.wikipedia.org/wiki/Program_Files), the directory name can be localized, so these paths may break on some systems. It's best to use Program Files environment variables instead.

Can Vuze be installed in both the `x86` and "regular"?

cleaners/vuze.xml
```diff
@@ -18,34 +18,91 @@
...
-    <action command="delete" search="glob" path="$APPDATA\Azureus\active\*.bak"/>
+    <!--
+    Was: (Tobias)
+    <action command="delete" search="glob" path="$$Profile$$\*.bak"/>
+    <action command="delete" search="glob" path="$$Profile$$\active\*.bak"/>
+    <action command="delete" search="glob" path="$$Profile$$\subs\*.vuze.bak"/>
+    -->
+    <!-- Is now: (Tobias) -->
+    <!-- Delete *.bak recursive: -->
+    <action command="delete" search="walk.files" path="$$Profile$$\" regex="\.[Bb][Aa][Kk]$"/>
```
az0: FYI, if there are hundreds of files and folders in the profile, this might be slow.

cleaners/vuze.xml
az0: You mean 32-bit Windows, Vuze is installed in the `c:\program files` diretory and on 64-bit Windows it's installed in `c:\program files(x86)`? If so, then the regular  `%ProgramFiles%` would work as long as BleachBit is 32-bits.

cleaners/vuze.xml
Tobias-B-Besemer: AFAIK the installer is 32/64-bit... So it's ever in \Program Files\Vuze...

Tobias-B-Besemer: Look: https://en.wikipedia.org/wiki/Environment_variable#Default_values
`%ProgramW6432%[20] | Yes | %ProgramW6432% (not supported, not replaced by any value) | %SystemDrive%\Program Files (only in 64-bit version)[20]`

Tobias-B-Besemer: OK, it seems to be possible to install in both folders...
https://www.filehorse.com/download-vuze-32/

Tobias-B-Besemer: As far as I understand it, if you want that I/we use environment variables, then you must solve this with some code.
https://stackoverflow.com/questions/1283664/python-get-wrong-value-for-os-environprogramfiles-on-64bit-vista
A better docu, but .NET: https://weblog.west-wind.com/posts/2019/Feb/05/Finding-the-ProgramFiles64-Folder-in-a-32-Bit-App

cleaners/vuze.xml
Tobias-B-Besemer: OK, but I should go recursive... No one must use this option, if he/she don't want...

Tobias-B-Besemer:
No, on 64-bit the default Vuze installer installs a 64-bit version into `c:\program files` and on 32-bit the installer installs a 32-bit version into `c:\program files`. Vuze is a Java app and comes as 32- & 64-bit.

cleaners/vuze.xml
Tobias-B-Besemer: No, on W64 Vuze gets installed into `c:\program files` (64-bit version) and on W32 Vuze get installed into `c:\program files` (32-bit version).

az0: That seems odd, so it would help to have a comment in the XML

cleaners/vuze.xml
az0: I am going to change this to the default multi-value variable $$ProgramFiles$$ . If you seen an issue with let, me know.

PR #475, opened 2019-03-11
miro.xml

Tobias-B-Besemer:


cleaners/miro.xml
```diff
@@ -39,5 +51,10 @@
     <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log"/>
     <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log.1"/>
     <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro_Download_Daemon_a.txt"/>
+    <!-- Polish Windows XP Support: -->
+    <action command="delete" search="file" path="$USERPROFILE\Ustawienia lokalne\Temp\Miro-downloader.log"/>
```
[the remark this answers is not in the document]
Tobias-B-Besemer: OK, will do!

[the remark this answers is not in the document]
Tobias-B-Besemer: I guess I had a better idea... I just use `$Temp`... ;-)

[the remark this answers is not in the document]
Tobias-B-Besemer: ...also adds support for Windows Vista and higher... ;-)

cleaners/miro.xml
```diff
@@ -18,26 +18,48 @@
...
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log.1"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro_Download_Daemon_a.txt"/>
+    <action command="delete" search="file" path="$Temp\Miro-downloader.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log.1"/>
+    <action command="delete" search="file" path="$Temp\Miro_Download_Daemon_a.txt"/>
+  </option>
+  <option id="temp">
+    <label>Temporary files</label>
```
az0: Lines 58-59 are translated. Good

cleaners/miro.xml
```diff
@@ -18,26 +18,48 @@
...
+    <description>Delete the temporary files</description>
+    <action command="delete" search="file" path="$Temp\Miro-downloader.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log.1"/>
```
az0: Lines 60-62 deletes three log files that look like the three deleted on lines 52-54. Would you agree lines 60-62 should be deleted?

cleaners/miro.xml
```diff
@@ -18,26 +18,48 @@
...
     <action command="delete" search="file" path="~/.miro/miro-log"/>
     <action command="delete" search="file" path="~/.miro/miro-log.1"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro-downloader.log"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro.log.1"/>
-    <action command="delete" search="file" path="$USERPROFILE\Local Settings\Temp\Miro_Download_Daemon_a.txt"/>
+    <action command="delete" search="file" path="$Temp\Miro-downloader.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log"/>
+    <action command="delete" search="file" path="$Temp\Miro.log.1"/>
+    <action command="delete" search="file" path="$Temp\Miro_Download_Daemon_a.txt"/>
```
az0: Line 55 deletes the same file (Miro_Download_Daemon_a.txt) deleted on line 63. Would you agree line 55 should be deleted?

cleaners/miro.xml
Tobias-B-Besemer: IMHO, no! One it's the option "temp", one is the option "log".

cleaners/miro.xml
Tobias-B-Besemer: I would let it...

cleaners/miro.xml
Tobias-B-Besemer: I do my best! ;-)

cleaners/miro.xml
az0: I think a good principle is the same action shouldn't be repeated in the same cleaner file (either within the same option or in a different option). First, if there are options labeled "log" and "temporary files," it implies chosing one doesn't do the other. If you think for Miro they are similar, you could eliminate the "logs" option and put all the actions under "temporary files."

Tobias-B-Besemer: I prefer to remove "Temporary files", again. I will do this.

PR #476, opened 2019-03-11
adobe_reader.xml

Tobias-B-Besemer:


az0:
Set to milestone BleachBit version 2.4 final (also 2.3 beta)

Tobias-B-Besemer:
Can you explain me why appveyor & travis-ci fail, again?

az0:
The CI tests failed because there is a hard-coded test in `TestCLI.py` specifically for Adobe Reader. If you want not to worry about it, I will change it later.

az0:
Thanks for your hard work on this. This is a big change that requires time for review and testing, so I pushed back the milestone to not further delay all the good changes ready for the next release. If you could break up the change into smaller pieces, that could speed it up.

Tobias-B-Besemer:
Thanks for the compliment! :)
No prob with the push back! All fine! :)
Will try to find time again and then break it into smaller pieces... but if you want to do me a favor, lets do it first with the installer. Was 4 weeks work full-time and I'm a bit proud of it... ;) I already made a first piece of it as a PR...

az0:
I delayed the milestone to focus on Python 3, so BleachBit is not get removed from Debian.

az0:
This PR looks like it has many lines (in an XML comment) that were maybe copied and pasted from Winapp2.ini. Please don't submit any lines from Winapp2.ini because it might create copyright and licensing issues.

Also, it would greatly help to get this merged if it were split up into simpler pieces that are easier to review.

PR #478, opened 2019-03-11
google_chrome.xml

Tobias-B-Besemer:


az0:
I targeted to 2.4 because this often-used cleaner deserves a thorough look

Tobias-B-Besemer:
I guess you will include it in BB v2.3beta already?

az0:
You asked about v2.3. After 2.2 stable is released, version 2.3 beta can start

Tobias-B-Besemer:
@az0, you should really make a review of this PR! I have already v2.3.0 in the pipe, but it includes already a lot of comments about the "universal code base"... So better to make (first) a review of this version...
But "universal code base" (forked from v2.3.0) will give you cleaners for all Chromium based browsers... Something even CCleaner don't have! ;-)

az0:
Thanks for your hard work on this. This is a big change that requires time for review and testing, so I pushed back the milestone to not further delay all the good changes ready for the next release. If you could break up the change into smaller pieces, that could speed it up.

Tobias-B-Besemer:
Yeah, we can work on this together...
State ATM is, I liked the idea to give all Chome- (just as Firefox-) based browser an identical CleanerML-Code-Base. If the code base gets always synced in future, this should speed up changes and bring support for a lot of (smaller) browser!
But I should have the OK from you to it...
...and maybe a "discussion" about how to do it... My first intention was to write a "universal" one for Google Chrome and then "just copy it" to the other browsers... Prob with it is, the cleaner will try to clean "old files" from older versions of Google Chrome that maybe never existed in the forks... but I think this shouldn't be a problem of performance and if the file(s) never existed in the fork, there's nothing the cleaner can delete (by mistake)... ;) ...so what do you think about this all and how far can I go this way that you still check in the PRs to it ???
Btw.: If you don't like that BB tries to delete files never existed in the fork, this would make IMHO the solution much more complicated! Because then you have no "universal code base" anymore, you would have more a "template" for the forks and then sync the changes between all forks (with each a other version of the template) will get complicated and you easily can loss overview of it... In comparison an "universal code base" should be (IMHO) relatively simple to add additions the forks made and this addition can be marked/highlighted in the code, that a cleaner editor know which part is Google Chrome based and which part is "Fork added".
Only solution I can find ATM for not deleting files never existed with a "universal template" would be to build the cleaner (code) like a timeline, clean each version after the other and cut older versions from the forks when they never were based on, but how have the docs about which file came up with which version and when they disappeared again and who is willing in spending time into the cleaner (codebase) and can maybe help me develop it ???

az0:
This has 487 changes. I delayed the milestone to focus on Python 3, so BleachBit is not get removed from Debian.

Tobias-B-Besemer:
@az0, I worked 3 months on the project full-time... Would be terrible if it wouldn't be many changes... ^^

az0:
I agree that I don't want to waste your time. The good is so far 81 of your commits have made it into the main repository plus 126 into the CleanerML repository, and there will be more.

az0:
I appreciate your work on this, Tobias. However, I am unable to review or merge these changes.

# First issue
With the original Google Chrome cleaner (before the changes in this PR), the stats for a preview are:
Disk space to be recovered: 1.11GB
Files to be deleted: 31250
Special operations: 32

With this cleaner in this PR, these are the stats for a preview:
Disk space to be recovered: 491.8MB
Files to be deleted: 27265

So this cleaner seems to cause a regression of ~8% of files and 55% of space. 

It's more complicated because probably some new files were cleaned, so the loss was greater. Also, systems vary. However, in any case, this PR clearly deletes fewer files.

# Second issue
At the end in XML comments, it seems many lines were copied and pasted from the Winapp2.ini project. Please do not include any lines from Winapp2.ini because this might cause copyright and licensing issues.

# Third issue
This pull request adds 476 lines to a file that was only 181 line. This is a huge change. It would greatly speed up the merging of this pull request to simplify it, so it can be reviewed. For example, if today I review the changes under the id=cache option, I can't merge just the lines I reviewed today. It's all or nothing, so I have to review all 476 lines. This makes it take dramatically longer to review and to merge. I am not asking for "microscopic" changes, but I am hoping for more manageable changes.

PR #464, opened 2019-03-07
internet_explorer

Tobias-B-Besemer:
Try to fix issue #460.

PR #485, opened 2019-03-13
BB2.2_IE_Polish-Support

Tobias-B-Besemer:


cleaners/internet_explorer.xml
```diff
@@ -41,19 +53,29 @@
...
   <option id="temporary_files">
     <label>Temporary files</label>
     <description>Delete the temporary files</description>
     <action command="delete" search="walk.files" path="%APPDATA%\Microsoft\Windows\IETldCache\"/>
     <action command="delete" search="walk.files" path="%USERPROFILE%\AppData\Local\Microsoft\Windows\Temporary Internet Files\"/>
-    <action command="delete" search="walk.files" path="%USERPROFILE%\Local Settings\Temporary Internet Files\"/>
     <action command="delete" search="walk.files" path="%LOCALAPPDATA%\Microsoft\Feeds Cache\"/>
+    <!-- Used e.g. in Windows XP (and maybe before on NT) - English-only-Support: -->
+    <action command="delete" search="walk.all" path="%USERPROFILE%\Local Settings\Temporary Internet Files\"/>
+    <action command="delete" search="glob" path="%USERPROFILE%\Local Settings\Temporary Internet Files\*"/>
```
az0: The glob search on line 75 will match all files directly in this folder, while the action above (line 74) already deleted all contents recursively, so line 75 will do nothing

cleaners/internet_explorer.xml
```diff
@@ -41,19 +53,29 @@
...
+    <!-- Used e.g. in Windows XP (and maybe before on NT) - some none-English-Support: -->
+    <!-- Polish Windows XP Support: -->
+    <action command="delete" search="walk.all" path="%USERPROFILE%\Ustawienia lokalne\Temporary Internet Files\"/>
+    <action command="delete" search="glob" path="%USERPROFILE%\Ustawienia lokalne\Temporary Internet Files\*"/>
```
az0: Same idea as my comment above but here with lines 78 and 79. I think you want to either remove the `*` asterisk from line 79 to delete the folder itself or delete the whole action

cleaners/internet_explorer.xml
Tobias-B-Besemer: I hope I get it now finally when and how to use `glob`! :-)

az0:
I am trying to release 2.2 soon, so I marked this for 2.4 to not delay version 2.2.

Tobias-B-Besemer:
It's just the Polish Fix you wanted for v2.2... If you don't want it anymore for v2.2, you can give this PR up, because the other PR have all Fixes including the Polish Fix...

az0:
I merged #464 instead of this one, so it has the fix.

I was hoping there would be an environment variable that contains the Polish path name, so the fix would work for other languages too, but I haven't found it.

Tobias-B-Besemer:
As far as I remember I clean in general more then before (all Windows versions)... and with using $$IELocalAppData$$ there should be additional cleaning in XP, too. Think this was the reason I created the variable... but some times ago I did it... so I don't have any desission anymore in mind... (...but hopefully I knew what I did... *lol*)

Tobias-B-Besemer:
...would be easily (now) to add more languages, but seems in all the years no one complained and it's WinXP support and we are on the point were you can plan to drop support for Win7 already... *lol* ...so guess no one cares... ^^

PR #489, opened 2019-03-13
filezilla.xml

Tobias-B-Besemer:


cleaners/filezilla.xml
```diff
@@ -12,24 +12,59 @@
 
     This program is distributed in the hope that it will be useful,
     but WITHOUT ANY WARRANTY; without even the implied warranty of
-    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+    MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
     GNU General Public License for more details.
 
     You should have received a copy of the GNU General Public License
-    along with this program.  If not, see <http://www.gnu.org/licenses/>.
+    along with this program. If not, see <http://www.gnu.org/licenses/>.
```
az0: I agree that one space after a period is good, but this is verbatim from the GNU license, which is in the root folder of the repository. While not a big deal either way, I suggest to leave it as accurate copy of that document.

az0:
I see one unresolved concern: https://github.com/bleachbit/bleachbit/pull/489/files#r265370101

PR #498, opened 2019-03-18
teamviewer.xml---new

Tobias-B-Besemer:


cleaners/teamviewer.xml
```diff
@@ -33,8 +33,8 @@
 <cleaner id="teamviewer" os="windows">
   <label>TeamViewer</label>
   <description>Remote control software</description>
-  <var name="profile">
-    <value os="windows">$AppData\TeamViewer</value>
+  <var name="Profile">
+    <value search="glob" os="windows">%AppData%\TeamViewer</value>
```
az0: `search="glob"` is unnecessary here because `%AppData%\TeamViewer` contains no wildcard

Tobias-B-Besemer: I remove now all `search="glob"` from `<value>` where it isn't necessary in my cleaners...

PR #504, opened 2019-03-22
mc.xml

Tobias-B-Besemer:


cleaners/mc.xml
```diff
@@ -33,15 +33,15 @@
...
   <var name="profile">
     <value os="windows">%UserProfile%\Midnight Commander</value>
     <value os="linux">~/.mc</value>
   </var>
   <option id="history">
     <label>History</label>
     <description>Delete the usage history</description>
-    <action command="delete" search="file" path="$$profile$$/filepos"/>
     <action command="delete" search="file" path="$$profile$$/history"/>
+    <action command="delete" search="file" path="$$profile$$/filepos"/>
```
az0: What is the importance of the order?

cleaners/mc.xml
```diff
@@ -33,15 +33,15 @@
 -->
 <cleaner id="midnightcommander">
   <label>Midnight Commander</label>
-  <description>File Manager</description>
+  <description>File manager</description>
```
az0: Yes, lowercase is consistent with the others

cleaners/mc.xml
Tobias-B-Besemer: None! I just reordered because I did the patch anyway... "history" on top because the option is "history"... No deeper sense of it!

cleaners/mc.xml
Tobias-B-Besemer: :-)

cleaners/mc.xml
Tobias-B-Besemer: Btw.: Reason for the PR was the "Fix @cleanerby history"...

Issue #335, opened 2018-04-18
button to update does not work (GTK+ 3)


az0:
When `update_dialog()` gets valid URLs, it does not pass them to `open_url()` because `open_url()` refers the URL as `False`.

Seen in ede855c883126c9b8ec7cf3d5fd5676e0a271498

az0:
I confirmed this is still a problem for branch gtk3 as of commit 3a3d923

````
Traceback (most recent call last):
  File "/bleachbit/bleachbit/Update.py", line 142, in <lambda>
    'clicked', lambda dummy: open_url(url, parent, False))
  File "/bleachbit/bleachbit/GuiBasic.py", line 171, in open_url
    ret = re.search('^http(s)?://([a-z.]+)', url)
  File "/usr/lib/python2.7/re.py", line 146, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer
````

Issue #334, opened 2018-04-18
--shred command line does not work with GTK+ 3 GUI


az0:
````
$ python bleachbit.py  --shred --gui COPYING 

Traceback (most recent call last):
  File "bleachbit.py", line 43, in <module>
    bleachbit.CLI.process_cmd_line()
  File "/bleachbit/bleachbit/CLI.py", line 215, in process_cmd_line
    app = bleachbit.GUI.Bleachbit(uac=not options.no_uac, shred_paths=args, exit=options.exit)
  File "/bleachbit/bleachbit/GUI.py", line 70, in __init__
    self.shred_paths(shred_paths)
AttributeError: 'Bleachbit' object has no attribute 'shred_paths'
````

Seen in Git commit ede855c883126c9b8ec7cf3d5fd5676e0a271498. 

It is basically the same on Windows 10 and Ubuntu 16.04

az0:
I confirmed this is still a problem for branch gtk3 as of commit 3a3d923

Issue #37, opened 2014-12-05
app-menu.ui strings are not included in bleachbit.pot


az0:
# original message

@moimael: Would you please look at including strings from data/app-menu.ui in bleachbit.pot using po/Makefile, so the strings can be translated?  

# requirements
- put strings to be translated into bleachbit.pot to upload to LaunchPad translation system
- use translated strings from the `.po` files for the strings in `data/app-menu.ui`

PR #534, opened 2019-04-15
Gtk3 fixes

mkhon:
Fix the following issues:
- #37 
- #335 
- #334 

For the latter it is not clear if
- we should ask for confirmation (currently we do)
- we should quit after shredding is done (currently we do)

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -54,6 +54,7 @@ def threaded(func):
 
 class Bleachbit(Gtk.Application):
     _window = None
+    _shred_paths = None
 
     def __init__(self, uac=True, shred_paths=None, exit=False):
         if uac and 'nt' == os.name and Windows.elevate_privileges():
@@ -67,7 +68,7 @@ class Bleachbit(Gtk.Application):
         GObject.threads_init()
 
         if shred_paths:
-            self.shred_paths(shred_paths)
+            self._shred_paths = shred_paths
             return
         if 'nt' == os.name:
             # BitDefender false positive.  BitDefender didn't mark BleachBit as infected or show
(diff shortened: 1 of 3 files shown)


po/Makefile
```diff
@@ -1,7 +1,7 @@
 APP = bleachbit
 POFILES = $(wildcard *.po)
 MOFILES = $(patsubst %.po,%.mo,$(POFILES))
-SRC = $(wildcard ../bleachbit/*.py)
+SRC = $(wildcard ../bleachbit/*.py) ../data/app-menu.ui
```
az0: Good

bleachbit/GUI.py
```diff
@@ -248,6 +249,9 @@ def do_activate(self):
         if not self._window:
             self._window = GUI(application=self,title=APP_NAME)
         self._window.present()
+        if self._shred_paths:
+            GUI.shred_paths(self._window, self._shred_paths)
+            GObject.idle_add(lambda: self.quit(), priority=GObject.PRIORITY_LOW)
```
az0: It runs as expected

bleachbit/GuiBasic.py
```diff
@@ -157,7 +157,7 @@ def message_dialog(parent, msg, mtype=Gtk.MessageType.ERROR, buttons=Gtk.Buttons
     return resp
 
 
-def open_url(self, action, url, parent_window=None, prompt=True):
+def open_url(url, parent_window=None, prompt=True):
```
az0: Runs as expected

az0:
You said two things were unclear. Asking for confirmation is good, and that's what you did. 

BleachBit version 2 with GTK 2 did not exit after `--gui --shred`, but with your change it, it does quit. It seems better the way you did it. Thanks

PR #548, opened 2019-04-24
deepscan.xml (Review wanted)

Tobias-B-Besemer:
Remove cache="false" + header


Tobias-B-Besemer:
Seems we both made the same patch... :-D
At least my one have a nice header with history info, so I push it anyway... ;-)

Btw.:
I work already on a other project, again. So I will spend less time per day in future, but therefor decided to work longer on the project (in weeks)... ^^ I want to finish all my sub-projects, but don't write any new on my to-dos-list, anymore. Then, after release of v3.0, I will work on new sub-projects, again. ^^

Issue #545, opened 2019-04-22
file chooser menu options are broken on Windows GTK 3


az0:
On Windows 10 (GTK 3.18) clicking first three menu options (Shred Files, Shred Folders, Wipe free disk space) causes these tracebacks. On Ubuntu 18.10 (GTK 3.24) it works fine.

````
Traceback (most recent call last):
  File "bleachbit\GUI.pyo", line 124, in cb_shred_file
  File "bleachbit\GuiBasic.pyo", line 94, in browse_files
AttributeError: 'GUI' object has no attribute 'window'
Traceback (most recent call last):
  File "bleachbit\GUI.pyo", line 135, in cb_shred_folder
  File "bleachbit\GuiBasic.pyo", line 43, in browse_folder
AttributeError: 'GUI' object has no attribute 'window'
Traceback (most recent call last):
  File "bleachbit\GUI.pyo", line 173, in cb_wipe_free_space
  File "bleachbit\GuiBasic.pyo", line 43, in browse_folder
AttributeError: 'GUI' object has no attribute 'window'
```

Issue #313, opened 2018-03-12
re-enable drag and drop for GTK+ 3


az0:
GTK2+ commit was b30558e63d3aca9184c6f71fea252866517fd1e9

az0:
I confirmed this is still necessary for branch gtk3 as of commit 3a3d9237000ae3129fe4a87ec48d0f78ad31b7cf

PR #549, opened 2019-04-26
Gtk3 fixes

mkhon:
- #313 - Drag and drop was actually working but it did not work if drop was performed on main textview (I fixed that)
- #545 
- shred by pasting from clipboard

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -101,6 +101,7 @@ class Bleachbit(Gtk.Application):
         # set up mappings between <attribute name="action"> in app-menu.ui and methods in this class
         actions = {'shredFiles': self.cb_shred_file,
                    'shredFolders': self.cb_shred_folder,
+                   'shredClipboard': self.cb_shred_clipboard,
                    'wipeFreeSpace': self.cb_wipe_free_space,
                    'shredQuit': self.cb_shred_quit,
                    'preferences': self.cb_preferences_dialog,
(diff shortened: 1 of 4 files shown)


az0:
The drag-and-drop and "shred from clipboard" work OK on Ubuntu 18.10. Thank you.

First, please would you change it so it does not close the window except when using `--shred --gui` from the command line? In other words, any of these should not close the window: drag and drop, shred files from clipboard, and the regular cleaning button on the menu bar. 

Second, the GUI no longer starts on Windows 10 with GTK 3.18. It throws warnings (which I have not seen before) and then quits.

````
> bleachbit.py

** (python.exe:8920): WARNING **: Trying to register gtype 'GMountMountFlags' as enum when in fact it is of type 'GFlags'

** (python.exe:8920): WARNING **: Trying to register gtype 'GDriveStartFlags' as enum when in fact it is of type 'GFlags'

** (python.exe:8920): WARNING **: Trying to register gtype 'GSocketMsgFlags' as enum when in fact it is of type 'GFlags'
````

mkhon:
1) Can you clarify about closing the window? It does not close the window when I do drag and drop or shred and also cleaning button - tested on Windows 10 (GTK 3.18) and Ubuntu 18.04 LTS

2) Again, I did all the implementation and testing on Windows 10 + GTK 3.18 and I do not see these messages on start.

mkhon:
Screen Shot 2019-04-29 at 12 31 34

mkhon:
May be the problem is that I was testing with --gui --no-uac - DND does not work for me if UAC is done, however the app still starts up properly.

az0:
You were right: rebooting Windows 10 fixed the issue that BleachBit did not start. It starts fine now.

Yes, Windows does not allow DND with elevated permissions, so that is why I added "shred paths from clipboard."

About the window closing: I see this on Ubuntu 18.10 but not on Windows 10. When I preview some items, it does not close. After cleaning anything, it closes.
Here is a video https://youtu.be/Pm35I0oG9zE

mkhon:
Hmm, it does not happen for me in Ubuntu 18.04 and Windows 10 + GTK 3.18. I will try to see what happens on 18.10.

Does it happen on bleachbit gtk3 branch? Can you run "git bisect" from bleachbit gitk3 HEAD (rev. d5ab76fd71698d3170e58ed53c3129f86dbd399c). None of my last three commits can theoretically cause this.

mkhon:
I can not reproduce automatic quite (after clean) on Ubuntu 18.10 as well.

az0:
The "closing" issue was my fault. Thanks for your patience.

PR #582, opened 2019-07-17
Re-order buttons in delete confirmation dialog

### LINKED ISSUE #436: Bleachbit bad design, almost like Boeing MCAS
Why am I comparing the Bleachbit bad design to the flawed Boeing MCAS that costs hundreds of lives recently? Read below...

Long story short, in the recent Boeing 737-MAX Lion Air crash one of the angle of attack sensors on the 737MAX crapped out, and the airplane mistakenly thought it was pitched at a much higher AoA than it actually was. They were also getting unreliable airspeed indications. Traditionally the older (NG) 737 variants did not have fly-by-wire technology, and autopilot could be overridden and turned off simply by putting manual pressure on the yoke. Suffice it to say, the MAX's MCAS system ( Maneuvering Characteristics Augmentation System ) overrode the pilot even AFTER the pilots turned autopilot completely OFF and took manual control right after takeoff when they realized something was wrong. It refused to give the pilots control back and nose-dived the aircraft into the ocean at over 400mph. The idea was Boeing wanted to learn from Airbus mistake after the Air France Flight 447 incident in which mid-Atlantic they flew into a storm, some pitot tubes frozen over, and fly-by-wire system degraded, giving the pilots full manual control on their side-stick (Airbus uses joystick unlike Boeing with steering column) and one of the pilots got disoriented and forgot the basics of flying 101 and pitched the nose up at 40 degrees and stalled the A330 from FL410 all the way smashing 200mph into the ocean in less than two minutes... So Boeing thought it was going to be smart and designed the Maneuvering Characteristics Augmentation System (MCAS) to detect for high AoA situations and automatically pitch the nose down and keep it pitched down until the AoA returned to stable levels. While in the case of 737-MAX Lion Air the AoA sensor itself was broken and Boeing MCAS didn't do cross check nor any redundancy checks and relied entirely one that single broken sensor to pitch the nose down and basically dive the aircraft into the water killing all on board, and the pilots never had a chance, though they thought they turned autopilot off... The worst part of this is that Boeing neglected to put this in the manual and didn't tell any of the pilots, and no pilots were trained on the MCAS nor did they have simulator exercises on how to fix a runaway MCAS situation...

Many pilots and those in the airline associations now have lost some trust/faith/respect in Boeing, their thinking is that if Boeing intentionally withheld info on MCAS, what else did Boeing put in the 737MAX that the pilots and airlines don't know about? 

What does this have to do with BleachBit and this posting about bad design? 

In short, Bleachbit puts the "delete" confirmation button on the WRONG/OPPOSITE side of the dialog box, against customary convention and counter-intuitive to muscle memory in the Windows platform. 

For example this morning I wanted to load a bunch of text files into my notepad++ that were on the desktop (Windows 10) and I had not yet put in my contacts and things were still a blur... On this particular file that I needed to use I accidentally clicked on Bleachbit's shredder (which came auto-installed with the bleachbit itself) in the context menu when I meant to click on the "Open with Notepad++" entry... Then I realized that I had misclicked and that it invoked Bleachbit instead of opening it with Notepad++. I immediately clicked on the right side button of the dialog box, thinking it meant to cancel or was a "no" and that the confirm or delete was on the left hand side, and yet I found out too late and only after the fact that Bleachbit intentionally and with absolutely no good reason nor cause swapped the location of the buttons and was directly responsible for permanently destroying my file and erasing my data making it unrecoverable...

mkhon:
Fixes #436

### DIFF
--- a/bleachbit/GuiBasic.py
+++ b/bleachbit/GuiBasic.py
@@ -134,8 +134,8 @@ def delete_confirmation_dialog(parent, mention_preview):
     dialog.get_content_area().pack_start(hbox, False, True, 0)
     dialog.get_content_area().set_spacing(10)
 
-    dialog.add_button(_('_Cancel'), Gtk.ResponseType.CANCEL)
     dialog.add_button(_('_Delete'), Gtk.ResponseType.ACCEPT)
+    dialog.add_button(_('_Cancel'), Gtk.ResponseType.CANCEL)
     dialog.set_default_response(Gtk.ResponseType.CANCEL)
 
     dialog.show_all()


az0:
Great, thanks

Issue #533, opened 2019-04-11
Encoding issue? Non-english chars don't display e.g. 'åäö'


Jon-guy30:
So I changed the "National settings" in the control panel in Windows briefly to US settings, and then I changed it back to the default. Now when bleachbit encounters characters such as "åäö" (Swedish vowels) it's replaced by the "Encoding error", you know the square "?" symbol. 

It's a cosmetic issue but it bothers me, I want it to display them correctly again, like it used to. 

I reinstalled it with Swedish language, clean installed it (removed bleachbit from appdata folder), reinstalled it again with English language, then I saw there was an ini file in appdata, but I don't think it would do anything.  

I hope you don't mind me asking, I don't know how to fix this. 

EDIT: Bleachbit portable works fine, those special characters are displayed correctly there.

EDIT 2: How odd. When I preview files to be deleted in the application itself it displays correctly. But when I do it through the context menu it shows the replacement character. 

Uuuuuuuuuh.

az0:
So it fails only when using the Context Menu, right? It works correctly when it is installed or portable, as long as you don't use the context menu?

Jon-guy30:
Correct. I have a second computer which is a laptop where I recently reinstalled Windows on, it has the same issue.

az0:
@Jon-guy30 
I have a contractor working on a fix for this and related issues. When the fix is ready, would you be willing to test it and let us know?

az0:
@Jon-guy30 

Would you please test a potential fix from here: [build 2.3.1148](https://ci.appveyor.com/project/az0/bleachbit/builds/26073328/artifacts) ? The fix is not yet merged, so it won't be in other builds for now, so please use this link for 2.3.1148.

az0:
@Jon-guy30 I'm hoping you can test the fix. Let me know

Tobias-B-Besemer:
@az0, as far as I understand it, this issue should be fixed... So close this issue, now ???

az0:
@Tobias-B-Besemer 
OK, I am closing for lack of response. I assume this was fixed in 3.0.

Issue #567, opened 2019-06-22
Error when enabling options that show a pop-up warning


ediowar:
Windows 8 SL 64 bit
Run bleachbit.exe as administrator. Impossible to mark many points. Here is what can be marked.
<a href='https://postimg.cc/1fMdCNz2' target='_blank'><img src='https://i.postimg.cc/1fMdCNz2/1.png' border='0' alt='1'/></a>
<a href='https://postimg.cc/8jT3mbTL' target='_blank'><img src='https://i.postimg.cc/8jT3mbTL/2.png' border='0' alt='2'/></a>
[bleachbit.exe.log](https://github.com/bleachbit/bleachbit/files/3317252/bleachbit.exe.log)
[BleachBit.ini.txt](https://github.com/bleachbit/bleachbit/files/3317253/BleachBit.ini.txt)

az0:
Under Firefox, are you able to check any of the individual options except for password? From the error attached, it seems you would get the error when (1) choosing Passwords under Firefox and (2) when choosing everything under Firefox

az0:
In BleachBit click the menu and then **Diagnostics**. Please copy that and paste it here. If you would like, you can skip all the rows that refer to directories. I am looking for information about your language settings and operating system.

ediowar:
invalid UTF-8
<a href='https://postimg.cc/5jHxJnwb' target='_blank'><img src='https://i.postimg.cc/5jHxJnwb/invalid-UTF-8.png' border='0' alt='invalid-UTF-8'/></a>
BleachBit version 2.3
Git revision d660bbad
GTK version 3.18.9
SQLite version 3.28.0
FSE = mbcs
locale.getdefaultlocale = ('ru_RU', 'cp1251')
platform.win32_ver[1]() = 6.2.9200
platform.platform = Windows-8-6.2.9200-SP0
platform.version = 6.2.9200
sys.version = 2.7.16 (v2.7.16:413a49145e, Mar  4 2019, 01:30:55) [MSC v.1500 32 bit (Intel)]
win32com.shell.shell.IsUserAnAdmin() = True

ediowar:
Took screenshots when choosing everything.

az0:
@ediowar I have a contractor working on a fix for this and related issues. When the fix is ready, would you be willing to test it and let us know?

az0:
@ediowar 


[...]

PR #586, opened 2019-07-18
Fix unicode errors when we are previewing or checking nodes.

rados:
About #533 I would like to say that I couldn't reproduce it but I made a fix that could help. Maybe we can give a version to the user and he could check.


rados:
Closing this PR and separating commits for #533 and #567 because #567 has been fixed and the user that waits for it would be able to enjoy the fix.

PR #592, opened 2019-07-24
Fix windows10 build

mkhon:
This also removes api-ms-win-* files from the distribution that exist on the Appveyor build hosts.
These files are API stubs and can not (and should not) be redistributed.


az0:
@mkhon  thank you

PR #595, opened 2019-07-31
Switch from os.walk to scandir.walk to improve performance

### LINKED ISSUE #512: Some cleaner take long to "load" at start, when option "only needed" is checked ("Vuze Case")
Some cleaner take long to "load" at start, when option "only needed" is checked ("Vuze Case")
My English... :-D

OK, I have now the "Vuze Case" also on BiglyBT and on my Windows Cleaner...

Is it really sure, that the check stops, after the first file, the cleaner finds ???

Can you display a bit more what the cleaner do in console for debugging ???

Whats about the case, I cleaned e.g. Vuze immediately before, then the cleaner can't find anything and have to run to the end, right?

Is maybe a < Detect >-Rule a solution for such scanners that have to scan a big amount of files ???

ppolewicz:
Resolves #512:

python 2.7 using `os.walk`:
```bash
$ time python bleachbit/CLI.py -p vuze.cache3
Disk space to be recovered: 0
Files to be deleted: 0

real	0m50,547s
user	0m0,834s
sys	0m0,929s
```

and using `scandir.walk`:
```bash
$ time python bleachbit/CLI.py -p vuze.cache3
Disk space to be recovered: 0
Files to be deleted: 0

real	0m0,755s
user	0m0,423s
sys	0m0,068s
```
tests ran on a VM@hdd with a flushed kernel cache on a flat directory with 20k files. The start time of bleachbit includes eager loading of all the rules, so if that was excluded, the performance boost would be even more profound.

**Before this is merged, a depedency from [scandir](https://pypi.org/project/scandir/) has to be added to the project.**

### DIFF
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -43,6 +43,8 @@ import subprocess
 import tempfile
 import time
 
+import scandir
+
 logger = logging.getLogger(__name__)
 
 if 'nt' == os.name:
@@ -164,7 +166,7 @@ def children_in_directory(top, list_directories=False):
             for pathname in children_in_directory(top_, list_directories):
                 yield pathname
         return
-    for (dirpath, dirnames, filenames) in os.walk(top, topdown=False):
+    for (dirpath, dirnames, filenames) in scandir.walk(top, topdown=False):
         if list_directories:
             for dirname in dirnames:
                 yield os.path.join(dirpath, dirname)


ppolewicz:
The test above has shown an improvement by a factor of 66.949 on 20k files, but I later ran the same test on 200k files to isolate it from startup time.
Result using version before this PR:
```bash
real	75m3,426s
user	0m5,790s
sys	0m13,213s
```
and after PR:
```bash
real	0m2,630s
user	0m2,083s
sys	0m0,123s
```
so the speedup is not ~67 times but ~1712 times, when considering flat directories.

Issue #589, opened 2019-07-19
v.2.2 portable on Win 7Pro x64: ImportError: DLL load failed


caganer:
BleachBit-2.2-portable.zip
MD5: 274661b26b8776db7da90abed4c6b32b

bleachbit.exe.log says

Traceback (most recent call last):
  File "bleachbit.py", line 44, in <module>
  File "bleachbit\__init__.pyo", line 353, in <module>
  File "bleachbit\Windows.pyo", line 59, in <module>
  File "win32gui.pyo", line 12, in <module>
  File "win32gui.pyo", line 10, in __load
ImportError: DLL load failed: The specified module could not be found.
Traceback (most recent call last):
  File "bleachbit.py", line 44, in <module>
  File "bleachbit\__init__.pyo", line 353, in <module>
  File "bleachbit\Windows.pyo", line 59, in <module>
  File "win32gui.pyo", line 12, in <module>
  File "win32gui.pyo", line 10, in __load
ImportError: DLL load failed: The specified module could not be found.

caganer:
Edit: in case it is not self-explanatory already from the actual log copied above, the error stated precluded launch so nothing was "done" before the error occurred, other than running the executable.

az0:
That's the right md5 sum.  Some of BleachBit is working, but 

Which antivirus/security software do you use? For testing purposes, could you temporarily disable it in case it is blocking the DLL load?

caganer:
No real time antivirus is installed on this Windows 7 Pro x64 laptop except for the bundled Windows Defender. Occasionally I manually run MalwareBytes AdwCleaner v7.4.0 and ClamWin v.0.99.4 but afaik neither is a permanently memory resident application.
To diagnose this issue I first added bleachbit.exe to Excluded Files and Folders in Windows Defender Options but it still would not run.  
I then disabled Windows Defender Real-Time Protection but same result. 
Last, I added the instal folder for the executable to Windows Defender's Excluded Files and Folders  but same result.

The log file as last updated by the executable is as follows:

Traceback (most recent call last):
  File "bleachbit.py", line 44, in <module>
  File "bleachbit\__init__.pyo", line 353, in <module>
  File "bleachbit\Windows.pyo", line 59, in <module>
  File "win32gui.pyo", line 12, in <module>
  File "win32gui.pyo", line 10, in __load
ImportError: DLL load failed: The specified module could not be found.
Traceback (most recent call last):
  File "bleachbit.py", line 44, in <module>
[...]

PR #598, opened 2019-08-02
Fix #567

### LINKED ISSUE #567: Error when enabling options that show a pop-up warning

rados:
The user that reported the issue had two problems - one with checking unicode three nodes and second that the main app menu showed utf-8 errors. The second problem has been fixed with commit with message "Gtk.Builder uses system dgettext() for "translatable" items." Here I am merging the fix for the checkboxes which has been already verified by the user. We can give him a build to verify both issues fixed.


az0:
I will take a look

az0:
@rados 
The user ediowar tested and replied
>Invalid UTF-8 fixed. Still errors:
Ошибка во время выполнения операции 'firefox': 'cannot import name _common'
Ошибка во время выполнения операции 'windows_explorer': ''ascii' codec can't encode characters in position 55-65: ordinal not in range(128)'
Ошибка во время выполнения операции 'skype': 'cannot import name _common'



Feel free to communicate directly with this users (and users on other issues), if you like

rados:
@az0 
Just pushed a fix for this error (which basically fixes #533 as well because it is the same issue): 

Ошибка во время выполнения операции 'windows_explorer': ''ascii' codec can't encode characters in position 55-65: ordinal not in range(128)'

For the  'cannot import name _common' part as far as my short check and research showed there could be something in the build itself. I can reproduce it from the custom build given to user but cannot reproduce it from source code. I know, I have to learn how to build locally and trace it, but does this time come?

az0:
The traceback for `cannot import name _common` come from what seems to be an unrelated part of the code. (It's the same DLL issue as #589.) Let's consider #567 fixed.

rados:
Okay, thanks. Just to add that I started to create a test for this part 'characters in position 55-65: ordinal not in range(128)' and will give it to you soon.

PR #603, opened 2019-08-05
Fix #451

### LINKED ISSUE #451: UnicodeDecodeError in Console when executing dism.exe
Code is:
`<action command="process" wait="true" cmd="dism.exe /online /cleanup-image /spsuperseded"/>`

Console says:
2019-02-26 06_12_03-bleachbit

And nothing gets saved in the log/debug file!

rados:
Tested on win 7 with Russian language settings.


az0:
Thanks, Radoslav! I will ask Tobias (who reported this) to test. 

Also, it would be nice to have a unit test to validate this fix and check for regressions.

az0:
Sorry if I sound like a "broken record" about unit tests. :) If I could easily reproduce this on my system, then I wouldn't wait for the user to confirm. The ideal case on the is the new unit test fails without the code fix, and the unit test passes with the fix. 

Not sure if this would work on English Windows systems (like AppVeyor), but you could write Unicode text to a file and then have run_external() run the Windows "type" command against the file.

az0:
Sorry if I sound like a "broken record" about unit tests. :) If I could easily reproduce this on my system, then I wouldn't wait for the user to confirm. The ideal case on the is the new unit test fails without the code fix, and the unit test passes with the fix. 

Not sure if this would work on English Windows systems (like AppVeyor), but you could write Unicode text to a file and then have run_external() run the Windows "type" command against the file.

rados:
Pushed an extension to the test_process unit test but it doesn't pass because mock.patch cannot find the Action module. I can achieve this locally with adding '.../bleachbit/bleachbit' to PYTHONPATH. Is this acceptable solution?

az0:
Changing `Action.General.run_external` to `bleachbit.Action.General.run_external` fixes the ImportError on Windows 10 and on Linux.

On Linux, the new test passes without the fix. On Windows, the test hangs until I hit ENTER. Then it terminates without neither a success nor an error.

rados:
Rebased this branch with latest master branch code because there were related changes. Will make new pull request.

PR #611, opened 2019-08-20
Fix #318

### LINKED ISSUE #318: cannot read winapp2.ini with BOM
Edited with Notepad on Windows 10

````
Error reading winapp2.ini cleaner 'C:\Users\x\AppData\Roaming\BleachBit\cleaners\winapp2.ini'
  File "\bleachbit\Winapp.py", line 399, in load_cleaners
    inicleaner = Winapp(pathname)
  File "\bleachbit\Winapp.py", line 150, in __init__
    self.parser.read(pathname)
  File "C:\Python27\lib\ConfigParser.py", line 305, in read
    self._read(fp, filename)
  File "C:\Python27\lib\ConfigParser.py", line 512, in _read
    raise MissingSectionHeaderError(fpname, lineno, line)
MissingSectionHeaderError: File contains no section headers.
file: C:\Users\x\AppData\Roaming\BleachBit\cleaners\winapp2.ini, line: 1
'\xef\xbb\xbf\n'
````

rados:
Here I fixed only the case of the utf8 BOM. If the ini has different BOM we will crash as before. So, please tell me if we need to support more BOM options.

Also as part of this fix I fixed the common.skipUnlessWindows decorator as it hadn't worked correctly. The tests decorated with it had seemed to pass but they hadn't been called at all.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -168,8 +168,6 @@ class Winapp:
             content = fp.read().decode('utf-8-sig').encode('utf8')
             self.parser.readfp(StringIO.StringIO(content))
 
-        #self.parser.read(pathname)
-
         self.re_detect = re.compile(r'^detect(\d+)?$')
         self.re_detectfile = re.compile(r'^detectfile(\d+)?$')
         self.re_excludekey = re.compile(r'^excludekey\d+$')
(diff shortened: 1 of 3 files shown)


az0:
I merged into branch rados-ms1

Issue #316, opened 2018-03-12
Re-enable strip and UPX optimizations for Windows with GTK+ 3


az0:

az0:
Fixed in 51ddfc2f

commit 469c2e2a
AppVeyor 2.3.1281
BleachBit-2.3-portable.zip (14,057,900 bytes)
BleachBit-2.3-setup-English.exe (10,973,619 bytes)
BleachBit-2.3-setup.exe (11,081,581 bytes)

commit 51ddfc2f  (upx branch)
AppVeyor build 2.3.1282
upx reduced size of the dist directory by 16,875,008 B from 33,939,596 B to 17,064,588 B in 57.9 s
BleachBit-2.3-portable.zip (12,491,213 bytes)
BleachBit-2.3-setup-English.exe (11,560,518 bytes)
BleachBit-2.3-setup.exe (11,665,918 bytes)

Tobias-B-Besemer:
az0, the EXE files got a bit bigger...
Do you know why?

az0:
The EXE installers have better compression than the ZIP portable file. However, applying compression to files that are already compressed can make the archive bigger. (You might see this also when zipping multimedia files such as .jpg.) What's more important is that the contents are smaller, and that's shown here (grabbed directly from the build logs). This is more important because the installer is typically used temporarily and then deleted, while the installation is more permanent.

2.3.1281
Final size of the dist folder: 33,827,044

2.3.1282
Final size of the dist folder: 16,952,200

Therefore, the improvement is significant: 16,874,844 bytes  (49.9%).

Tobias-B-Besemer:
If you want use (Windows) features like "Repair", or "Add/Remove Component" the installer must persist... However, we talk about MBs in times of TB HDDs... so it doesn't really matter IMHO, if later some MBs more or less are used on the HDD...
Anyway, we want to deliver "perfect" (as possible) software, right?
So two other questions to this, too...
Do you pack the headers of the EXEs like bleachbit.exe and bleachbit_console.exe, too?
And I saw there are lots of "Themes" in the BB folder, but BB seems to use just one (on Win7)... Are they for Win10, or are Themes for e.g. Linux and other OSes included, too?

az0:
>Do you pack the headers of the EXEs like bleachbit.exe and bleachbit_console.exe, too?

The command is `upx.exe --best --crp-ms=999999 --nrv2e`

[...]

PR #621, opened 2019-09-10
Suggested new parameters for 7-Zip

Tobias-B-Besemer:


Tobias-B-Besemer:
@az0, guess before I make more tries with changes I'm not sure if they right and push more CI, you should finish this patch...

Tobias-B-Besemer:
If the build script would work with my changes...

2019-09-10 07_38_03-I__work_Cleaner_CI_BleachBit-2 3 1257-portable_BleachBit-Portable

2019-09-10 07_38_57-I__work_Cleaner_CI

az0:
library: py2exe expects a .zip, and I'm not sure how deep you'd have to dig in py2exe to change that. It would involve shipping 7zip, and that might eat up the benefit of 1345-811KB.

Portable: Many people don't have 7zip installed, and it's important to reduce the friction of installation to keep users happy. Also, once upx executable compression is re-enabled (it was disabled initially for GTK 3), the difference between .zip and .7z could considerably shrink.

Tobias-B-Besemer:
What's about "7ZIP and UPX" at http://www.py2exe.org/index.cgi/BetterCompression? May this help or you doing it already?

Tobias-B-Besemer:
Is Py2Exe only needed for the Portable Version? Here's a sample to create with Py2Exe and NSIS and 7-Zip a file http://www.py2exe.org/index.cgi/SingleFileExecutable...

az0:
Like shown in the article, I already use 7zip to repack the library.zip

Like shown in the article, I used UPX to pack Windows executables. It was temporarily turned off for GTK 3 development. Last I checked, UPX with GTK 3 caused an application startup error. Here is a ticket https://github.com/bleachbit/bleachbit/issues/316

py2exe is used on Windows for both the installed and portable versions. All files are identical except for the NSIS installer and one `.ini` file.

PR #623, opened 2019-09-10
directx.xml

Tobias-B-Besemer:
- Deletes DirectX MRUs

### DIFF
--- a/cleaners/directx.xml
+++ b/cleaners/directx.xml
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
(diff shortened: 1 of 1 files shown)


abitrolly:
@Tobias-B-Besemer looks like https://github.com/bleachbit/winapp2.ini is the better place to add this as it is purely Windows specific. Maybe to https://github.com/MoscaDotTo/Winapp2 I don't know.

@az0 how are these files in https://github.com/bleachbit/winapp2.ini used. Are they shipped with Windows installer? Merged with upstream during release? How to test them with development BleachBit version from source?

PR #647, opened 2019-10-19
fix for [Errno 27] when wiping free space on fat fs

jamazi:
This will fix infinite loop when wiping free space on a fat partition that has more than 4gb freespace.


az0:
@jamazi Thanks for the fix.

Which system did you test it on? I tested the old BleachBit and your change on Windows Version 10.0.18362.145 with an empty 10GB FAT32 partition, and both BleachBit versions seemed to work the same.

jamazi:
Debian 10

az0:
Yes, it also seems better on Ubuntu 19.04 with 10GB loopback FAT32 file system. Thanks

PR #649, opened 2019-10-22
Add installation instructions and prerequisites to README

Huluti:
In order to simplify installation and building from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,9 +4,7 @@ BleachBit cleans files to free disk space and to maintain privacy.
 
 ## Installation instructions
 
-### Linux
-
-[![Packaging status](https://repology.org/badge/tiny-repos/bleachbit.svg)](https://repology.org/project/bleachbit/versions)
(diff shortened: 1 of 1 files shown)


az0:
How about instead a link to the [installation documentation](https://docs.bleachbit.org/doc/install-on-linux.html)? It's more thorough. Also, the documentation directly contradicts what you wrote. The documentation states:

>Installing using apt from the distribution’s repositories is not recommended because they tend to carry old releases of BleachBit.

But I would be willing to expand the documentation to include the system repository methods, like you wrote, so the user has a choice.

(The documentation is also a Git repository, by the way.)

Huluti:
As you want, even though I think it's easier for users to use packages than go through github ...

az0:
The instructions on the docs site suggest a package. The difference is it's
often newer

Huluti:
No problem ;)

PR #689, opened 2019-11-11
Playing with the 7-Zip parameters

Tobias-B-Besemer:


Tobias-B-Besemer:
BleachBit-3.0.1.1362-portable.zip had: 12,679,488 bytes

BleachBit-3.0.1.1363-portable.zip has: 12,654,463 bytes

Is this an interesting change ???

az0:
Percentage wise 25K bytes is small, but it's progress, so yes.

Did you want to make more changes or do more experiments, or are you done?

Tobias-B-Besemer:
Mhhmmm... I can also test other settings... but I would suggest you check this PR in and then we compare against this settings in future.

PR #692, opened 2019-11-12
Activate more output of NSIS by compile

Tobias-B-Besemer:
Q&D - I hope it works!

Addresses: https://github.com/bleachbit/wishlist/issues/69


Tobias-B-Besemer:
K, not more to see ATM in AppVoyer...

I use the /V3 Switch...
https://nsis.sourceforge.io/Docs/Chapter3.html
..."The /V switch followed by a number between 0 and 4 will set the verbosity of output accordingly. 0=no output, 1=errors only, 2=warnings and errors, 3=info, warnings, and errors, 4=all output."...
...but maybe it's to much for fetching and analiyze... then you have to go down on /V2 again...

PR #699, opened 2019-11-13
Fix checkout

Tobias-B-Besemer:
Guess a prob coming from the introduction of .gitattributes


az0:
Is this change 100% line encodings, or are there other changes?

Tobias-B-Besemer:
First it appeared, I copied a complete new downloaded version over it to go sure! This should be commit "Just to go sure... UAC.nsh"... I fear, that I finally should fix this line-ending/checkout prob upstream... :-/

Tobias-B-Besemer:
I reported this prob now upstream here... https://github.com/Drizin/NsisMultiUser/pull/31 ...as a PR with the ".gitattributes" file I have created. I hope he will check it in.

PR #700, opened 2019-11-13
vlc.xml: running type="exe" from os="linux" to os="unix"

Tobias-B-Besemer:
Have a look at https://github.com/bleachbit/bleachbit/pull/699 because of the second file.


Tobias-B-Besemer:
Uh! Saw ATM that this cleaner requires maybe some clarification...
There is no os="linux, windows" in it...
First question: Is set os="linux, windows" possible ???
(Note: I should not the answer in the doc...)
ATM this cleaner should get executed on UNIX systems, too...
...but <var name="Profile"> & <var name="Config-File"> have only os="windows" & os="linux"...
...so I think this cleaner gets shown and executed on UNIX systems, too, but do nothing... Right ???
Think I/we should fix that...

Tobias-B-Besemer:
@az0, have in this PR UAC.nsh in... 'cause of .gitattributes...
I will close this PR now... make a new one later...
...but I think we should discuss the UNIX prob before here...

az0:
> There is no os="linux, windows" in it...
> First question: Is set os="linux, windows" possible ???

I don't see that it's necessary. The way it's coded now already does what I think you want. 

> ATM this cleaner should get executed on UNIX systems, too...

You mean chage linux to unix? OK.

Tobias-B-Besemer:
> You mean chage linux to unix? OK.

Thought this was what I did... ^^

...OK, not everywhere! So I make a new PR...

PR #702, opened 2019-11-14
NSIS.MUI2.MultiUser-Defines (Review wanted!)

Tobias-B-Besemer:


Tobias-B-Besemer:
I try to fix this PR, too...

Tobias-B-Besemer:
Looks IMHO now good, too! :)

Tobias-B-Besemer:
This is what I always had fear for!
When I make multiple PRs against a single file, seems the changes can overwrite each other!
:-/

windows/bleachbit.nsi
```diff
@@ -56,6 +58,44 @@
...
+; !define VERSION "2.3"
+; "VERSION" already defined!
+!define PROGEXE "${prodname}.exe"
+; !define COMPANY_NAME "BleachBit"
+; "COMPANY_NAME" already defined!
+
+; An option (MULTIUSER_INSTALLMODE_ALLOW_BOTH_INSTALLATIONS) defines whether simultaneous per-user
+; and per-machine installations on the same machine are allowed. If set to disallow, the installer
+; alaways requires elevation when there's per-machine installation in order to remove it first.
+!define MULTIUSER_INSTALLMODE_ALLOW_BOTH_INSTALLATIONS 1
```
az0: Please set ALLOW BOTH to 0 to reduce confusion and complexity. It seems extremely rare that anyone would want two installations. Even when I do testing, I don't have two installations. If I want to test two versions, I might use portable versions.

windows/bleachbit.nsi
```diff
@@ -56,6 +58,44 @@
...
+; MULTIUSER_INSTALLMODE_DEFAULT_ALLUSERS:
+; 0 or 1, (only available if MULTIUSER_INSTALLMODE_ALLOW_ELEVATION = 1 and there are 0 or 2 installations
+; on the system) when running as user and is set to 1, per-machine installation is pre-selected, otherwise
+; per-user installation.
+!define MULTIUSER_INSTALLMODE_DEFAULT_ALLUSERS 1
+
+; MULTIUSER_INSTALLMODE_DEFAULT_CURRENTUSER:
+; 0 or 1, (only available if there are 0 or 2 installations on the system) when running as admin and
+; is set to 1, per-user installation is pre-selected, otherwise per-machine installation.
+!define MULTIUSER_INSTALLMODE_DEFAULT_CURRENTUSER 0
```
az0: Note to self: besides the ALLOW_BOTH change, CURRENTUSER=0 is the only functional change in this commit. This change is fine.

az0:
Please see my comment about ALLOW_BOTH. Also, there is a merge conflict. Otherwise, this PR looks good to merge.

windows/bleachbit.nsi
Tobias-B-Besemer: But as far as I understand the description, in this case you can't install system wide when a user have BB already installed.
I thought e.g. in a corporate environment the admin comes always first and the user that have installed BB before must later deinstall second version... Was just my view as an old admin...

windows/bleachbit.nsi
Tobias-B-Besemer: Hey, I documented the code, too! ^^

Tobias-B-Besemer:
Where can I see the reason for the conflict ???

windows/bleachbit.nsi
az0: @Tobias-B-Besemer 
In a corporate environment, I trust that the admin would keep BleachBit up to date. If not, the user could use a portable version.  However, generally users shouldn't try to change what admins do. The admin might not appreciate it. Any any case, I think this situation would be rare.

windows/bleachbit.nsi
az0: I appreciate the documentation. :)  I consider documentation a non-functional change because it doesn't directly affect the function of the application.

az0:
> Where can I see the reason for the conflict ???

Does [this link](
https://github.com/bleachbit/bleachbit/pull/702/conflicts) work for you? 

I see two conflicts. The first is scriptversion/scriptdate around line 20. The second is a comment near line 163 above `!addplugindir /x86-unicode ".\NsisPluginsUnicode\"`

Tobias-B-Besemer:
No, link doesn't work, but doesn't matter!
The prob is, that the PRs overwrite the changes of each other! :-/
We must in future land one PR after each other...

az0:
You might have seen this build error

```
Processing config: C:\Program Files (x86)\NSIS\nsisconf.nsh
Processing script file: "windows\bleachbit.nsi" (ACP)
Plugin directories:
  C:\Program Files (x86)\NSIS\Plugins\x86-ansi
2019-12-01 15:56:11 - ERROR - Plugin not found, cannot call UAC::_
Error in macro UAC_RunElevated on macroline 1
Error in macro MULTIUSER_FUNCTION_INSTALLMODEPAGE on macroline 129
Error in macro MULTIUSER_PAGE on macroline 14
Error in macro MULTIUSER_PAGE_INSTALLMODE on macroline 1
Error in script "windows\bleachbit.nsi" on line 186 -- aborting creation process
2019-12-01 15:56:11 - ERROR - windows\BleachBit-3.0.1.1429-setup-English.exe not found
Command exited with code 1
```

Tobias-B-Besemer:
But the error wasn't there!
Git seems to have some difficulties, when multiple PRs target a single file...
I had this fear before...
...and this was the reason, that some PR got so big, because else I had to wait till you checked in, because I need to have the version last checked in.

Tobias-B-Besemer:
Think you can make review again and land this dragon... ^^

az0:
Is this covered by the PR #721?

Tobias-B-Besemer:
Yes!

PR #715, opened 2019-11-25
DNF autoremove support

klslz:
Introducing DNF autoremove support. 
Orphaned packages will be removed. The feature requires root privileges. 
There's no safety belt for the user. Packages, as being offered by dnf autoremove, 
will be removed without interaction. 
This cleaner relies 100% on the reliability of the dnf package manager to not brick the system by autoremoving crucial packages.

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -675,5 +675,19 @@ def yum_clean():
     new_size = FileUtilities.getsizedir('/var/cache/yum')
     return old_size - new_size
 
+def dnf_clean():
+    """Run 'dnf clean all' and return size in bytes recovered"""
+    if os.path.exists('/var/run/dnf.pid'):
+        msg = _(
+            "%s cannot be cleaned because it is currently running.  Close it, and try again.") % "Dnf"
+        raise RuntimeError(msg)
+
+    old_size = FileUtilities.getsizedir('/var/cache/dnf')
+    args = ['--enablerepo=*', 'clean', 'all']
+    invalid = ['You need to be root', 'Cannot remove rpmdb file']
+    run_cleaner_cmd('dnf', args, '^unused regex$', invalid)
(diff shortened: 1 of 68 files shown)


bleachbit/Unix.py
```diff
@@ -687,7 +687,42 @@ def dnf_clean():
     invalid = ['You need to be root', 'Cannot remove rpmdb file']
     run_cleaner_cmd('dnf', args, '^unused regex$', invalid)
     new_size = FileUtilities.getsizedir('/var/cache/dnf')
+
     return old_size - new_size
 
+units = {"B": 1, "k": 10**3, "M": 10**6, "G": 10**9}
+
+def parseSize(size):
```
az0: It would be nice to have a unit test, even a simple one, for `parseSize()`

cleaners/dnf.xml
```diff
@@ -27,4 +27,9 @@
     <description>Delete the cache</description>
     <action command="dnf.clean_all"/>
   </option>
+  <option id="autoremove">
+    <label translate="false">autoremove</label>
```
az0: I'm going to add a warning here

bleachbit/Unix.py
```diff
@@ -687,7 +687,42 @@ def dnf_clean():
...
+           "%s cannot be cleaned because it is currently running.  Close it, and try again.") % "Dnf"
+       raise RuntimeError(msg)
+    cmd = ['dnf', '-y', 'autoremove']
+    process = subprocess.Popen(cmd, stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+    freed_bytes = 0
+    while True:
+       line = process.stdout.readline().replace("\n", "")
+       if 'Error: This command has to be run under the root user.' == line:
+          raise RuntimeError('dnf autoremove >> requires root permissions')
+          break
```
az0: `break` is unreachable after `raise`

PR #722, opened 2019-12-01
7-Zip.Compression

Tobias-B-Besemer:


az0:
Build 3.0.1.1438 (with this change)
BleachBit-3.0.1.1438-portable.zip: 12,748,059 bytes

Build 3.0.1.1430 (recent build without this change)
BleachBit-3.0.1.1430-portable.zip: 12,681,476 bytes

So the change increased the size

Tobias-B-Besemer:
I saw this, too! But I needed a break! ;)
At leasr we can now fine tune the portable.zip separate from the library.zip! ;)

Tobias-B-Besemer:
Short explanation to the size (later more to it): Seems the compression method PPMd saves a lot of space with small files (library.zip), but waste a lot by big archives (portable.zip)... I need to do some new tests/comparison and play with the parameters of 7-Zip... ;)

az0:
While it's possible to do a bit better, I expect there's little margin to improve the portable archive because most of its contents (e.g., executables, images, and the library.zip archive) are already compressed.

Tobias-B-Besemer:
Yes, but if I look at your past work, it looks like you are a perfectionist... I be one, too! And sometimes I like a bit modding and tweaking... ;) ...so why not try to find out, whats really the perfect parameters for this archives? ;)

Tobias-B-Besemer:
Btw.: I tested some of my different library.zip test files with the portable version... And yes, it's true, the portable version doesn't work with a e.g. with Deflate64 compressed archive!
As we both chated about this and tested it, we made a big mistake! We just had a look on the archives that they can be opened with the build in Win10 unzip method!
But py2exe doesn't use the build in unzip methode of Win10, it have it's own decompression algorithm! And it seems, that this algorithm is only able to decompress Deflate compressed archives!

What I have now done (and it's easy to see in the PR, because I needed to change just some lines for it) is the following:
Till now the compression parameters got defind by the variable "SZ_OPTS" one time for both archives. I copied now the variable to SZ_OPTS_LIB for the library who can only use Deflate and to SZ_OPTS_APP for the pirtable app. After changing which cariable is used in which case, we are now able to set the parameter for each archive, even if library.zip is still present. :)

And btw.: I guess I found the reason why the EXE headers of the NSIS installers don't get compressed, ATM. The path to the packer seems to be incomplete! But this must/should be changed in the NSI file and the PRs of this file should get merged serial and I have already 3 PRs open for it, so I decided, to don't write one for it, yet.

az0:
I'm closing this because the current PR makes the files bigger. If you want to try again later, that's fine

Tobias-B-Besemer:
Andrew, I wasn't just finished, yet... :-/

az0:
Reopened

az0:
The PR splits the single SZ_OPTS variable into SZ_OPTS_LIB (for library.zip) and SZ_OPTS_APP (for portable.zip), then switches the outer portable archive from Deflate to PPMd compression.

Splitting them has merit, but portable.zip contains already compress payloads (exe, dll, png, library.zip). Also, PPMd works best on text and structured data but poorly on binaries and compressed data, which shows here.

Because there isn't a compression improvement and no changes lately, I am closing this. 

Thank you, Tobias, for your work on this and other things, and sorry for not keeping up with all your contributions.

PR #729, opened 2019-12-09
Ignore junctions on Windows

mkhon:


assarbad:
@az0: Just to remark. What this checks are not _just_ junction points, but reparse points, so the name of the function is misleading. Also, if I am comprehending this commit correctly, this will break BleachBit on all my computers and perhaps other peoples' as well.

## Junction Points < Reparse Points

It may seem nitpicky, but the difference matters and I'll outline why. I was just about to file a ticket regarding an issue caused by following a particular type of **reparse points** myself. After seeing this ticket, I decided to piggyback onto it instead.

So the name of the file system entity we're talking about is a **reparse point**. The junction points introduced with Windows 2000 are a strict subset of reparse points. In fact they were the first widely available (albeit not widely used) type of reparse points along with volume mount points.

Alas, the symbolic links introduced with Vista are also a type of reparse point and they can be subdivided into symbolic links to files and to folders. The reason why this matters is the following: unlike junction points, symbolic links can point to remote network shares and so on. That's why I wanted to file a ticket, actually (for now I will simply exclude the respective location). In my case I have a few (directory) symbolic links to remote network shares. This means BleachBit tries to clean (or scans) loads of folders I don't want it to peek into.

Quite some time ago I wrote a tool to inspect the various types of reparse points. You can find it [here](https://bitbucket.org/assarbad/looklink/). It is available under very liberal terms and so its code could be included. But you could also simply look at what it's doing and redo that in Python. I'll have to see whether I find time to spend on BleachBit. I certainly should after only having _used_ it and donated money but never time (which in my own FLOSS projects is always the actual bottleneck ;)).

## Why the distinction matters

The reason this distinction above matters and why this commit will break BleachBit on all of my machines is that junction points are a neat and convenient way to redirect data to another location. In my case I have - for over a decade - stored the browser profiles of those browsers I use as well as email clients and often some other things inside an encrypted container. Of course this has the downside of these programs not being usable while the encrypted container isn't mounted, but that's a small price to pay for the benefit.

Anyway, just to give you a hint, the following folders on my systems are actually junction points, pointing inside an encrypted container:

* `%LOCALAPPDATA%\Mozilla`
* `%APPDATA%\Mozilla`
* `%LOCALAPPDATA%\Opera`
* `%APPDATA%\Opera`
* `%LOCALAPPDATA%\Vivaldi`
* `%APPDATA%\Vivaldi`

... and _I certainly want BleachBit to consider these_ (well, I think Vivaldi isn't currently covered perhaps, but that's not the point).

I have other (non-browser) programs for which I am using the same method.

Another use case of junction points is to "extend" the disk space while leaving (most) programs oblivious about it. I've used that numerous times when bulky programs would not have fit on `C:` without major headache, but junction points helped to redirect to a place where disk space was available in abundance. I also used this to "redirect" individual folders (such as "Documents") to another drive for the same reason. By the way, while this wasn't the official method of redirecting well-known folders from within your profile, this one didn't break with one of the recent Windows 10 upgrades, while the upgrade went ahead and wiped out the profile folders when you had used the official method.

I think I am not the only one to have discovered these use cases or others in the last twenty years (i.e. since junction points got introduced in Windows 2000).

## Why the distinction also matters

Superficially this could be the end of the story. Alas, it is not. The deduplication feature available on Windows Server uses reparse points. DFS (Distributed File System) uses reparse points.

Reparse points are a generic facility. A reparse point has a so-called [reparse point tag](https://docs.microsoft.com/en-us/windows/win32/fileio/reparse-point-tags). This one encodes for example whether the the reparse point type is owned by Microsoft or a third party. A GUID is then used to identify the exact type of reparse point (roughly this means that the tag is a fast way for a particular file system filter driver to figure out if it is responsible for the reparse point handled by the I/O manager and then use the GUID to take the _appropriate_ action; that is a single driver may be responsible for multiple types of reparse points).

To tinker with reparse points and other file system entities [this project](https://github.com/googleprojectzero/symboliclink-testing-tools) and [this project](https://github.com/ladislav-zezula/FileTest) may be of interest.

## Conclusion

The intent of the commit is perfectly fine, but I think this calls for some setting in the preferences rather than blankly ignoring _all_ reparse points, given the variety of reparse points in existence.

az0:
@assarbad 

Would you please download [build 1447](https://ci.bleachbit.org/?prefix=dl/3.0.1.1447/) and run a preview to check how it handles your browser profiles?

assarbad:
@az0 just tried it and this build can "see through" the junction points instead of ignoring them. Thanks (and sorry for the late response, I missed the notification email).

az0:
@assarbad 
Thanks for checking and writing back

PR #787, opened 2020-02-19
optimization for Windows theme

TraistaRafael:
Windows 10 theme optimization. Css minifyed and png compressed using optipng.
image


az0:
I am not sure of an easy way to resolve this. The problem is this Git base from which you started. I squashed your first commit and then applied it, but in your branch, it's not squashed, so there are conflicting commits in the two histories before the theme changes. I think you should update your fork before applying the theme changes.

Alternatively, if there is a script that applies the optimizations, then you could send me the script to run on my end.

az0:
Also, please use the master branch. I merged py3 (include your Windows 10 theme work) into the master branch, and I will get rid of the py3 branch

az0:
I am closing this in favor of the newer https://github.com/bleachbit/bleachbit/pull/789

PR #871, opened 2020-04-13
Update Action.py

adam083b:
Splitting cmd arguments didn't work as expected, I suggest using the shlex module as more appropriate.


az0:
Would you please add a unit test to demonstrate the condition in which this is needed?

echo083:
Hello,

It's just to parse correctly spaces and other special character.
Let me describe my case which led me to make the proposal.

See below my custom_home cleaner:
[custom.xml.zip](https://github.com/bleachbit/bleachbit/files/4523355/custom.xml.zip)

If you use  args = self.cmd.split(' ') the cmd="rm $HOME/.config/ReText\ project/ReText.conf" will not be parsed correctly and will ignore the space character of the folder.

If you run  bleachbit -c custom.clear_retext_configuration you will get the following output
Command: rm /home/echo/.config/ReText\ project/ReText.conf
Return code: 1
Stdout: 
Stderr: rm: cannot remove '/home/echo/.config/ReText\': No such file or directory
rm: cannot remove 'project/ReText.conf': No such file or directory

But with shlex the space character is correctly interpreted:
Run external command: rm /home/echo/.config/ReText\ project/ReText.conf 0 
Disk space recovered: 0
Files deleted: 0
Special operations: 1

Thanks

az0:
[BleachBit 4.9.1 alpha](https://www.bleachbit.org/news/bleachbit-491-alpha) was released with this fix.

PR #952, opened 2020-06-24
Add gspawn-win32-helper.exe for opening external URLs (About dialog)

### LINKED ISSUE #918: Website button fails
Button in "about BleachBit" that's supposed to link to the BleachBit website doesn't work, screenshot below ...
Annotation 2020-05-10 132313

mkhon:
Fixes #918

### DIFF
--- a/windows/setup_py2exe.py
+++ b/windows/setup_py2exe.py
@@ -217,6 +217,12 @@ def build():
     if not os.path.exists('dist'):
         os.makedirs('dist')
 
+    logger.info('Copying GTK helpers')
+    shutil.copyfile(os.path.join(GTK_DIR, 'gspawn-win32-helper.exe'),
+                    os.path.join('dist', 'gspawn-win32-helper.exe'))
+    shutil.copyfile(os.path.join(GTK_DIR, 'gspawn-win32-helper-console.exe'),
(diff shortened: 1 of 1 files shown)


az0:
I downloaded `BleachBit-4.0.1.1650-portable.zip` and unpacked on a machine without a development environment. I confirmed it has the file `gspawn-win32-helper.exe`, but I still get the error "No application is registered as handling this file."

mkhon:
Original report was about "Failed to execute helper program" - this PR fixes this error.
I'll check "No application is registered as handling this file" error.

mkhon:
I think the PR can be safely merged as it fixes original issue.

mkhon:
I noticed two things:
- one more helper (gspawn-win32-helper-console.exe) is required for bleachbit_console.exe
- "No application is registered as handling this file" occurs when default web browser is Microsoft Edge. If I set it to Chrome or Internet Explorer (11) urls are opened as expected - this seems to be a bug in GTK+ and I'd suggest to open a separate bug about this (which will hopefully be fixed with GTK+ update)

az0:
Yes, I see the same thing: with Edge, there is "no application is registered." With Google Chrome, it opens the URL.

Issue #928, opened 2020-05-26
Wrong language


tmlz:
I specifically downloaded the English installer, but it still installed another language and I am not able to change it to English.

az0:
The [BleachBit FAQ](https://docs.bleachbit.org/doc/frequently-asked-questions.html) states
>To force BleachBit to always show in English, download and install the English installer instead of the regular installer. If using the regular (non-English-only) installer, uncheck the option to install translations.

>The language chosen in the BleachBit installer does not affect the BleachBit application: it affects only the language of the installer.



Does this help?

tmlz:
No, I did download the English installer

az0:
This may be related to a [known bug ](https://github.com/bleachbit/bleachbit/issues/864)in the English installer.

For now, one option is to install the regular installer (not the English installer) and uncheck the translations.

A second option is to delete the localization (i.e., translation files). I believe they are in `c:\program files (x86)\bleachbit\share\locale`. If you don't want them, it's safe to just delete them, and the application should start in English.

az0:
@tmlz 

Would you please test the fix? It will be published in 10 minutes at https://ci.bleachbit.org/ . Please test version 4.0.1.1668 or newer

IgorArnaut:
How about I change language inside the program???

az0:
> How about I change language inside the program???

Please don't reply to an old, closed issue and change the subject ("necroposting"). 

If for a bug, enhancement, or feature: Instead, please search GitHub for the same open issue or create your own new issue ticket.

For technical support: use [discussions](https://github.com/bleachbit/bleachbit/discussions)

Issue #864, opened 2020-04-10
English installer installs localizations


az0:
The installer `BleachBit-3.9.1.1623-setup-English.exe` installed 126 files worth 1.59MB in `C:\Program Files (x86)\BleachBit\share\locale`

The English installer is supposed to be smaller for people who don't want this, but the size is about the same.

BleachBit-3.9.1.1623-setup-English.exe (12,617,676 bytes)
BleachBit-3.9.1.1623-setup.exe (12,729,819 bytes)

Gitoffthelawn:
I recommend using BleachBit to remove these unwanted files. ;)

az0:
This issue is the same as https://github.com/bleachbit/bleachbit/issues/770

>All kidding aside, where are these other builds located?

Look for bleeding-edge builds at https://ci.bleachbit.org/ . Sometimes there are updates more than once daily.

PR #959, opened 2020-06-25
Do not include "share/locale" into Core component

### LINKED ISSUE #770: Installer installs BleachBit with unwanted translation
The installer offered to install BleachBit with English language, and althoguh I'm not a native English speaker, I was fine with that because I assume every time that the translation is faulty or not complete.
After installation, the program hardly opened after a few minutes (Process Hacker showed high IO usage (~30-40 MB/s) without any disk usage), and it had the translation for my language active. I did not want that, I requested English to be used. Also, the translation is not complete, it has a lot of missing pieces.

I have no idea why was it installed in my language. My system (Windows 10 1903) has English as the default language, but it has my language installed too.
I feel like I need to mention that I have my language set up as the default for non unicode programs, and also that the input language is set to my language.

mkhon:
Fixes #770, #864, #928

### DIFF
--- a/windows/bleachbit.nsi
+++ b/windows/bleachbit.nsi
@@ -308,7 +308,7 @@ Section Core (Required)
     SectionIn RO
 
     SetOutPath $INSTDIR
-    File /r "..\dist\*.*"
+    File /r /x "locale" "..\dist\*.*"
     File "..\COPYING"
 
     SetOutPath "$INSTDIR\share\"
--- a/windows/setup_py2exe.py
+++ b/windows/setup_py2exe.py
@@ -563,7 +563,7 @@ def package_installer(nsi_path=r'windows\bleachbit.nsi'):
         # Was:
         # nsis('/DNoTranslations',
         # Now: Compression gets now done in NSIS file!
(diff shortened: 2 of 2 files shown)


az0:
# Installation test
After installing BleachBit-4.0.1.1667-setup-English.exe, there is no share/locale folder: **good**

After installing BleachBit-4.0.1.1667-setup.exe with translations component unchecked, there is no share/locale folder: **good**

After installing BleachBit-4.0.1.1667-setup.exe with translations component checked, there is share/locale folder: **good**


# Size
FIle sizes for build 1667 with this fix
BleachBit-4.0.1.1667-setup-English.exe (12,358,886 bytes)
BleachBit-4.0.1.1667-setup.exe (12,758,610 bytes)
Difference=399,724

File sizes without this fix
BleachBit-4.0.1.1666-setup-English.exe (12,649,162 bytes).
BleachBit-4.0.1.1666-setup.exe (12,761,974 bytes)
Difference=112,812

So fix made an improvement of about 286,912 bytes compressed, so that **saved 2% compressed**.

When installed, size of folder is 126 files with 1.89MB out of 18.5MB for the whole application, so that **saved 10% installed**.

Issue #943, opened 2020-06-14
Wholeregex matches any expression


Silur:
# How to reproduce

Make a custom cleaner under `~/.config/bleachbit/cleaners/test.xml:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<cleaner id="testscanner">
  <label translate="false">Test scanner</label>
  <description>Demonstrate regex bug</description>
  <option id="invalid">
    <label translate="false">invalid path</label>
    <description>this regex should not match anything</description>
    <action command="delete" search="deep" wholeregex="definitelynotvalid" />
  </option>
</cleaner>
```
Running this cleaner will match every file the deep scanner runs trough:
```
[ continuation ]
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/deref/Deref.t.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/deref/trait.DerefMut.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/deref/DerefMut.t.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/drop/trait.Drop.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/drop/Drop.t.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/arith/trait.Rem.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/arith/SubAssign.t.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/arith/Neg.t.html
Delete 4.1kB /home/silur/.rustup/toolchains/nightly-x86_64-unknown-linux-gnu/share/doc/rust/html/core/ops/arith/RemAssign.t.html
[...]

PR #965, opened 2020-06-29
DeepScan: implement shred and wholeregex

### LINKED ISSUE #321: deep scan does not shred

mkhon:
- Implement nregex, wholeregex, nwholeregex
- Implement command="shred"
- Case-insensitive search on Windows

Fixes #321, #943

### DIFF
--- a/bleachbit/Action.py
+++ b/bleachbit/Action.py
@@ -23,20 +23,16 @@
 Actions that perform cleaning
 """
 
-from bleachbit import Command, FileUtilities, General, Special
-from bleachbit import _
+from bleachbit import Command, FileUtilities, General, Special, DeepScan
+from bleachbit import _, fs_scan_re_flags
 
 import glob
 import logging
 import os
 import re
-
-
 if 'posix' == os.name:
-    re_flags = 0
     from bleachbit import Unix
-else:
-    re_flags = re.IGNORECASE
+
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


az0:
I see the refactoring, unit tests, and improvement with case sensitivity. Thank you, Max.

Test A: ok

In the GUI preferences, I enabled the overwrite option and cleaned "temporary files" under "deep scan." I tested both the PR and master branch (f8c3453e113e09facf2f0c83396e84c463b572bc).

In both cases, it *did* shred because variable `do_shred` was set to True in `delete()` function, so issue #321 was misinformed.


Test B: ok

I reset the configuration `bleachbit.ini`. Then, I ran the command line `python3 bleachbit.py --overwrite --clean deepscan.tmp`.  I tested both the PR and master branch.

In both cases, it *did* shred because variable `do_shred` was set to False in `delete()` function, so issue #321 was misinformed.


Test C: good

I edited `deepscan.xml` to change `command="delete"` to shred. In the master branch, there was a NotImplementedError, and in the PR, it *did* shred, so this was an improvement.


Test D: good

On Ubuntu, I tested case sensivity. I made files called `ppt2345.tmp` and `ppt2345.TMP`. With both branches, it matched only the former.


Test E: good

On Windows 10, I tested case sensitivity. I made files called `ppt1000.tmp` and `ppt1001.TMP`. With master, it matched only the former (case sensitive). With the PR, it matched both (case insensitive). This is progress.

PR #975, opened 2020-07-11
fixing-issue-944

devnet1985:
Its  work on windows 10 without any change.


az0:
Thank you for trying, but this is not helpful, so I am closing it

PR #989, opened 2020-08-21
Unnecessary import loop

flyingbot91:
Despite python only imports a module the first time (caches the next ones) I think there is no point in including the _import_ statement inside the _for_ loop.

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -657,9 +657,9 @@ def get_apt_size():
 
 def get_globs_size(paths):
     """Get the cumulative size (in bytes) of a list of globs"""
+    from glob import iglob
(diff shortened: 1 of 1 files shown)


az0:
Yes, it looks odd inside the loop. There's also an import at the module level, so how about this 316019f147429010e8f4c21ff944304583d71711

If these kind of changes interest you, see the site https://lgtm.com/projects/g/bleachbit/bleachbit/ or run `make lint`

flyingbot91:
@az0  https://github.com/bleachbit/bleachbit/commit/316019f147429010e8f4c21ff944304583d71711 seems ok to me.

I'll take a look at your link. Thanks!

Issue #856, opened 2020-04-05
 ResourceWarning: unclosed <socket.socket....


az0:
Seen in the AppVeyor [log for build 1618](https://ci.appveyor.com/project/az0/bleachbit/builds/31965878#L1509)

Git revision 27546b5d9b3f56769d2e3e12772ae254d229da01

```
Unit test for class Chaff ... Downloading https://sourceforge.net/projects/bleachbit/files/chaff/clinton_subject_model.json.bz2/download to C:\Users\appveyor\AppData\Local\Temp\1\bleachbit-chaffdiy812zn\subject.json.bz2
c:\python34\lib\socket.py:451: ResourceWarning: unclosed <socket.socket fd=28, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.0.3', 1103), raddr=('69.65.16.142', 443)>
  self._sock = None
Downloading https://sourceforge.net/projects/bleachbit/files/chaff/clinton_content_model.json.bz2/download to C:\Users\appveyor\AppData\Local\Temp\1\bleachbit-chaffdiy812zn\content.json.bz2
c:\python34\lib\site-packages\cryptography\hazmat\_oid.py:20: ResourceWarning: unclosed <socket.socket fd=740, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.0.3', 1102), raddr=('216.105.38.13', 443)>
  for node in nodes:
c:\python34\lib\site-packages\cryptography\hazmat\_oid.py:20: ResourceWarning: unclosed <socket.socket fd=336, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.0.3', 1101), raddr=('216.105.38.13', 443)>
  for node in nodes:
c:\python34\lib\socket.py:451: ResourceWarning: unclosed <socket.socket fd=1100, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('10.0.0.3', 1106), raddr=('172.110.242.68', 443)>
  self._sock = None
Downloading https://sourceforge.net/projects/bleachbit/files/chaff/2600_model.json.bz2/download to C:\Users\appveyor\AppData\Local\Temp\1\bleachbit-chaffdiy812zn\ts.json.bz2
[...]

PR #1032, opened 2020-10-21
Fix unclosed socket in issue #856

br-z0:


br-z0:
I'll work on this one.

PR #1041, opened 2020-11-03
Switching to Python 3.8 for Windows

### LINKED ISSUE #848: Update Python 3 in Windows installer
For BleachBit 4.0, we updated Python 2.7 to Python 3.4. This was major progress and allowed a quicker release by keeping py2exe. However, Python 3.4 is EOL, and py2exe does not support Python 3.5+. The goal of this issue is to update the Windows installer to use a new version of Python---preferably Python 3.8.

Candidates for replacing py2exe include pyInstaller and cx_Freeze.

A related issue is the [GTK installer on Sourceforge](https://sourceforge.net/projects/pygobjectwin32/files/) also does not support Python 3.5+. An alternative is to install the msys32 environment and its package mingw-w64-i686-gtk3 .

abitrolly:
Fixes #848.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,8 +26,8 @@ install:
   - '%PYTHON_HOME%/Scripts/pip.exe install requests'
   - '%PYTHON_HOME%/Scripts/pip.exe install plyer'
   - '%PYTHON_HOME%/Scripts/pip.exe install win-unicode-console'
-  - C:\msys64\usr\bin\pacman -Suy
-  - C:\msys64\usr\bin\pacman -S mingw-w64-x86_64-gtk3 mingw-w64-x86_64-python3 mingw-w64-x86_64-python3-gobject
+  - C:\msys64\usr\bin\pacman --noconfirm -Suy
+  - C:\msys64\usr\bin\pacman --noconfirm -S mingw-w64-x86_64-gtk3 mingw-w64-x86_64-python3 mingw-w64-x86_64-python3-gobject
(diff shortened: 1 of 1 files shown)


az0:
@abitrolly 
Thank you for trying. Please address the following error seentin the AppVeyor log

 `2020-11-03 06:42:44 - ERROR - C:\Python38\Lib\site-packages\gnome\ not found`

abitrolly:
@az0 found this project through UpWork. Might need a confirmation there before putting a lot of effort into migration.

abitrolly:
Because currently used GTK libs for Windows are Python 3.4 only, the further work requires finding the replacement for that, which means getting the list of all files 

* [ ] Get the list of files currently found and packaged by `py2exe`
  * http://www.py2exe.org/index.cgi/FAQ#How_does_py2exe_decide_which_modules_you_need.3F
  * Compare the list from command line scan with what actually happens during AppVeyour build
  * Unpack the installer and compare files from there

* [ ] Try run BleachBit with alternative libs
  * MSYS2
  * Artifacts from https://github.com/wingtk/gvsbuild project (https://github.com/wingtk/gvsbuild/issues/377)

az0:
I appreciate the start here, but I am closing for now because of inactivity. There is still a need for this, though. 

General note to anyone who may be reading, please discuss with me before starting on the problem.  It's non-trivial.

kenrap:
@az0 
> General note to anyone who may be reading, please discuss with me before starting on the problem. It's non-trivial.

I need to do this anyway, so yes, I want to discuss this with you. Since I use a VFIO Windows virtual machine for gaming, I might as well use it for testing Bleachbit too. Although, I recently upgraded it from Win10 to Win11... So, I don't know if that would be a problem?

Issue #1059, opened 2020-11-30
BleachBit-4.1.1.1770 Beta Error: firefox.url_history


sperglord8008s:
**Log of print out** 

Error: firefox.url_history: Function: Clean file: C:\Users\Usern\AppData\Roaming\Mozilla\Firefox\Profiles\bmso88xa.default-release-1587222501240\favicons.sqlite
Traceback (most recent call last):
  File "bleachbit\FileUtilities.pyo", line 440, in execute_sqlite3
sqlite3.OperationalError: unrecognized token: "'javascript:(function(){l=[].map.call(document.querySelectorAll("span.ytd-thumbnail-overlay-time-status-renderer"),function(e){l=e.innerHTML.trim().split(":").map(function(t){return parseInt(t)}).reduce(function(p,c){return p*60+c})"

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bleachbit\Command.pyo", line 160, in execute
  File "bleachbit\Special.pyo", line 344, in delete_mozilla_favicons
  File "bleachbit\FileUtilities.pyo", line 448, in execute_sqlite3
sqlite3.OperationalError: unrecognized token: "'javascript:(function(){l=[].map.call(document.querySelectorAll("span.ytd-thumbnail-overlay-time-status-renderer"),function(e){l=e.innerHTML.trim().split(":").map(function(t){return parseInt(t)}).reduce(function(p,c){return p*60+c})": C:\Users\User\AppData\Roaming\Mozilla\Firefox\Profiles\bmso88xa.default-release-1587222501240\favicons.sqlite

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "bleachbit\Worker.pyo", line 87, in execute
  File "bleachbit\Command.pyo", line 162, in execute
AttributeError: 'OperationalError' object has no attribute 'message'

**Diagnostics Screen -** 
BleachBit version 4.1.1
Build number 1770
GTK version 3.18.9
GTK theme = Adwaita
GTK icon theme = Adwaita
GTK prefer dark theme = True
SQLite version 3.33.0
local_cleaners_dir = None
locale_dir = C:\Program Files (x86)\BleachBit\share\locale\
options_dir = C:\Users\User\AppData\Roaming\BleachBit
personal_cleaners_dir = C:\Users\User\AppData\Roaming\BleachBit\cleaners
system_cleaners_dir = C:\Program Files (x86)\BleachBit\share\cleaners\
[...]

Issue #1058, opened 2020-11-27
Error in 4.1.1.1770 when cleaning Firefox bookmarks


Ivan75:
The same amount of deleted data with the same error shown multiple times, shown on the first SS

The second SS shows the results after restarting the system.

![Screenshot (27)](https://user-images.githubusercontent.com/39677610/100488760-6f364f00-3110-11eb-953f-c11a0de9fd31.png)
![Screenshot (28)](https://user-images.githubusercontent.com/39677610/100488765-75c4c680-3110-11eb-9acc-5a677b3ea208.png)

rados:
@Ivan75 May I ask you to test something, just delete your tubecast bookmark manually and try again to run Bleachbit app. 'javascript:location.href="tubecast:link="+location.href;'

Thanks.

Ivan75:
@rados 

Thanks, I did it, it helped with firefox but some errors remained.

az0:
@Ivan75 

Thanks for running that test. The new error seems unrelated to the Firefox bookmark problem, so let's consider that a separate issue/ticket.

Would you please run another test? Add the YouTube JavaScript bookmark that you removed earlier. Install [build 1778](https://ci.appveyor.com/project/az0/bleachbit/builds/36702913/artifacts), and clean Firefox -  URL History. The `OperationalError` message should not appear.

Ivan75:
Thanks for replying, I can't make another test for the next few days, my PC is in repair.
Ivan

az0:
> Thanks for replying, I can't make another test for the next few days, my PC is in repair.
@Ivan75 
I'm just following up. Did you get your PC back? I hope it's better :)

PR #1062, opened 2020-12-07
Fix for #1058 and #1059

rados:


abitrolly:
I am concerned that if those errors are caused by improper quoting, this might hide a security bug in BleachBit.

az0:
@rados 
Thanks for the quick fix, but I got this error

```
delete from moz_pages_w_icons where (page_url not in ('https://support.mozilla.org/en-US/products/firefox','https://support.mozilla.org/en-US/kb/customize-firefox-controls-buttons-and-toolbars?utm_source=firefox-browser&utm_medium=default-bookmarks&utm_campaign=customize','https://www.mozilla.org/en-US/contribute/','https://www.mozilla.org/en-US/about/','http://www.ubuntu.com/','http://wiki.ubuntu.com/','https://answers.launchpad.net/ubuntu/+addquestion','http://www.debian.org/','https://www.mozilla.org/en-US/firefox/central/','https://www.bleachbit.org//','javascript:(function(d){d.PwnBkmkVer=3;d.body.appendChild(d.createElement('script')).src='https://deturl.com/ld.php?'+1*new Date;})(document);'));
Error: firefox.url_history: Function: Clean file: /home/z/.mozilla/firefox/6rmkiwj3.default-release//favicons.sqlite
  File "/home/z/software/bleachbit/bleachbit/FileUtilities.py", line 444, in execute_sqlite3
    cursor.execute(cmd)
sqlite3.OperationalError: near "script": syntax error
```

The problem bookmark is `javascript:(function(d){d.PwnBkmkVer=3;d.body.appendChild(d.createElement('script')).src='https://deturl.com/ld.php?'+1*new Date;})(document);`

Issue #780, opened 2020-02-13
Linting issues flagged on AppStream metainfo


terrycloth:
There seem to be some minor issues:
```
• tag-missing           : <content_rating> required [use https://odrs.gnome.org/oars]
• style-invalid         : <caption> is too long [BleachBit 3.0 in dark mode showing discovered Google Chrome files];longest allowed is 50 chars
```

Otherwise looks good, thanks!

_Originally posted by @AsciiWolf in https://github.com/bleachbit/bleachbit/pull/747#issuecomment-583737534_

terrycloth:
@AsciiWolf, what linter threw these warnings for you?

[`<content_rating>`](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-content_rating) wasn't mentioned in the AppStream "quickstart" so I hadn't noticed this was an option. I think it could be safely omitted, since the spec calls it optional, and BleachBit is a tool, not a game, and so is very unlikely to include blood and gore... But I guess the single tag -- `<content_rating type="oars-1.0" />` -- generated by the OARS questionnaire wouldn't be difficult to add either.

On the other hand, the fact that the official URL (<https://odrs.gnome.org/oars>) redirects to <https://hughsie.github.io/oars/> makes me wonder if this is _actually_ supposed to be an official part of the specification yet... The AppStream documentation for `<content_rating>` doesn't even list all the relevant child tags and attributes.

I'm also curious what the 50-character limit for screenshot captions is based on. The AppStream documentation recommends that captions be "short" but doesn't recommend a hard limit. Is 64 characters going to get truncated? Because if the text is not chopped off, I would personally be inclined to leave the flagged caption as it is --- BleachBit's UI is focused around text, without graphical visualizations of the data it finds, so I think this slightly more detailed caption is useful for explaining to new users what exactly they're looking at in the screenshot.

AsciiWolf:
@terrycloth I used standard `appstream-util validate` cli tool.

terrycloth:
@AsciiWolf Do you see the screenshot caption getting truncated in your software center? Or is there documentation stating that that's what will happen? Because looking again [at the documentation about the `<caption>` element](https://www.freedesktop.org/software/appstream/docs/chap-Quickstart.html), it now says the maximum caption length should actually be _180_ characters, and we're well within _that_ limit.

terrycloth:
[...]

PR #1087, opened 2021-01-07
Add OARS content rating, shorter screenshot caption

terrycloth:
Solves Issue #780.

...Technically, the screenshot taken from the website is for BleachBit 3.2, but the UI hasn't changed much in v4.x.

### DIFF
--- a/org.bleachbit.BleachBit.metainfo.xml
+++ b/org.bleachbit.BleachBit.metainfo.xml
@@ -18,7 +18,7 @@
   </description>
   <screenshots>
     <screenshot type="default">
-      <caption>BleachBit in dark mode, listing found junk files</caption>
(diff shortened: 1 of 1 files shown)


abitrolly:
Isn't the dark mode the default?

terrycloth:
> Isn't the dark mode the default?

Oh, right, yes, good point.

abitrolly:
@terrycloth my though was that if the dark mode is the default, then the title can be shortened even more. )

terrycloth:
My reasoning had been that since dark UI themes are relatively uncommon outside terminals and code editors, it was worth pointing that it's the default instead of... some random custom theme, or whatever. But I don't have strong feelings about the exact wording, as long as it's short enough to close issue #780.

PR #1093, opened 2021-01-10
Better fix for removing firefox favicons with attaching places db in …

rados:
…the query.

Here I introduced the right fix for deleting firefox favicons. We won't warry anymore for escaping quotes or dealing with semicolons in bookmarted urls. Just attaching the places.sqlite did the trick.

Now, should I go back to related tickets and ask the users to test this?

### DIFF
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -422,7 +422,7 @@ def exe_exists(pathname):
         return exists_in_path(pathname)
 
 
-def execute_sqlite3(path, cmds, cmds_as_list=None):
+def execute_sqlite3(path, cmds):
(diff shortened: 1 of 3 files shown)


az0:
> Now, should I go back to related tickets and ask the users to test this?

Sure, within a few minutes, it will post to https://ci.bleachbit.org/ as build 4.2.0.1804

rados:
> Sure, within a few minutes, it will post to https://ci.bleachbit.org/ as build 4.2.0.1804

Okay, I just informed our Windows users. Do we a have Linux builds somewhere since we have a couple of Linux users as well?

abitrolly:
`git clone ...` works for me on Linux. Although I am not ready to clean up Firefox yet.

PR #1104, opened 2021-01-28
When building for Windows add segoe ui in dist/share/fonts folder and…

rados:
… remove windows fonts dir from fonts.conf file.

### DIFF
--- a/windows/setup_py2exe.py
+++ b/windows/setup_py2exe.py
@@ -26,6 +26,7 @@ import subprocess
 import sys
 import time
 import win_unicode_console
+import xml.dom.minidom
 
 setup_encoding = sys.stdout.encoding
 win_unicode_console.enable()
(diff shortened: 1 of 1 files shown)


az0:
@rados 

I took a quick look. The idea is promising. Thank you

Including the fonts in the distributed files increases the size and can create other issues. Instead of being packaged in the installer/portable file, could the fonts be copied during installation via NSIS? For portable mode, maybe the fonts could be copied during the application's first start before initializing GTK.

abitrolly:
I like the idea of using own fonts dir to speed up caching of fonts by dealing only with the fonts that are used.

At some point I think it could be useful on Linux as well. Maybe we can move the font management into BleachBit itself, to make the font dir configurable, and copy fonts that people select there. That sounds like redoing the font-cache job, which we wouldn't have to do if it did well.

PR #1106, opened 2021-01-30
Fix Edge browser's exe name in the respective cleaner.

rados:


az0:
@rados 
Thank you. :+1:  In the future please cross reference the issue that the PR fixes by including the issue number. 😊

Issue #1035, opened 2020-10-26
TypeError: 'int' object is not subscriptable in Windows Wipe


Surfride10:
I am receiving the errors below with 4.1.0 attempting to delete app crash reports in WER in .\programdata...  target is 2012r2

Thanks!

~David

Delete 105.4kB C:\ProgramData\Microsoft\Windows\WER\ReportQueue\NonCritical_7.9.9600.19810_535d16b2204d8b274f9fc255741fd7bce31c71_00000000_cab_0e352a93\ReportingEvents.log
Error: system.logs: Command to delete C:\ProgramData\Microsoft\Windows\WER\ReportQueue\NonCritical_7.9.9600.19810_535d16b2204d8b274f9fc255741fd7bce31c71_00000000_cab_0e352a93\WindowsUpdate.log
Traceback (most recent call last):
  File "C:\projects\bleachbit\bleachbit\Worker.py", line 87, in execute
  File "C:\projects\bleachbit\bleachbit\Command.py", line 82, in execute
  File "C:\projects\bleachbit\bleachbit\FileUtilities.py", line 350, in delete
  File "C:\projects\bleachbit\bleachbit\FileUtilities.py", line 789, in wipe_contents
  File "C:\projects\bleachbit\bleachbit\WindowsWipe.py", line 968, in file_wipe
  File "C:\projects\bleachbit\bleachbit\WindowsWipe.py", line 281, in choose_if_bridged
  File "C:\projects\bleachbit\bleachbit\WindowsWipe.py", line 280, in <listcomp>
  File "C:\projects\bleachbit\bleachbit\WindowsWipe.py", line 226, in extents_a_minus_b
TypeError: 'int' object is not subscriptable
Delete 3.1kB C:\ProgramData\Microsoft\Windows\WER\ReportQueue\NonCritical_7.9.9600.19810_dd1e2b8f33e5e14571ee70ee48ef48f93e43ea59_00000000_cab_0bf86315\Report.wer
Error: system.logs: Command to delete C:\ProgramData\Microsoft\Windows\WER\ReportQueue\NonCritical_7.9.9600.19810_dd1e2b8f33e5e14571ee70ee48ef48f93e43ea59_00000000_cab_0bf86315\ReportingEvents.log
Traceback (most recent call last):
  File "C:\projects\bleachbit\bleachbit\Worker.py", line 87, in execute
  File "C:\projects\bleachbit\bleachbit\Command.py", line 82, in execute
[...]

Issue #1051, opened 2020-11-11
When wiping free disk space, only one 4GB file is written to the 8GB thumb drive and then deleted


ediowar:
```
clean_operation('_gui'), options = '[]'
Wiping free disk space erases remnants of files that were deleted without shredding. It does not free up space.Wiping path: G:\
Creating new, temporary file for wiping free space.
Wrote 1 files and 4,294,967,296 bytes in 1,003 seconds at 4.28 MB/s
Overwrite free disk space G:\ 0 

Disk space recovered: 0
Files deleted: 0
Special operations: 1
elapsed time: 1014 seconds
```
Windows 8 SL (64-bit)
BleachBit version 4.1.1-portable
Build number 1766
![Drive properties](https://user-images.githubusercontent.com/44164186/98851848-e361c900-2467-11eb-8389-4920bdfac3f2.png)

az0:
Good information, ediowar.

This seems like a regression related to the FAT32 file limit.

Marvin0401:
Yes, it seems like a regression related to the FAT32 file limit. Let me fix it soon.

Marvin0401:
Hi ediowar, I've fixed this issue. Could you test with this fork "https://github.com/Marvin0401/bleachbit.git"? If you need to get help from me about this issue, please feel free to let me know. Thanks

ediowar:
@Marvin0401 For testing on Windows, I need a portable version.

ediowar:
@Marvin0401 for testing on Windows, I need a portable version.

ediowar:
@Marvin0401, for testing on Windows, I need a portable version.

az0:
@MarvinQC
Submit a PR, and AppVeyor will automatically build and publish binaries for
users to test.

Marvin0401:
@ediowar , Sorry for late response. Please check it with this PR.
[BleachBit-4.1.1-portable-20201215-Marvin.zip](https://github.com/bleachbit/bleachbit/files/5696220/BleachBit-4.1.1-portable-20201215-Marvin.zip)

Marvin0401:
@az0 , AppVeyor is working well for building, but I've got failed to deploy S3 due to access_key_id.

az0:
@Marvin0401 

> @az0 , AppVeyor is working well for building, 

[...]

PR #1110, opened 2021-02-03
Fixed the issue #1035

Marvin0401:


abitrolly:
Would be nice to have more descriptive commit messages to read `git log` without opening GitHub.

bleachbit/FileUtilities.py
```diff
@@ -80,6 +80,36 @@ def open_files_linux():
...
+            path = os.sep.join(splitpath[:i]) + os.sep
+            if path in partitions:
+                return partitions[path]
+            
+            path = os.sep.join(splitpath[:i])
+            if path in partitions:
+                return partitions[path]
+
+        return ("unknown", "none")
+    except ImportError:
```
az0, line 108: How about movin the `except` right after `import` on line 88? It would be easier to read, and there's no way to throw an ImportError from lines 90-105.

bleachbit/FileUtilities.py
```diff
@@ -924,12 +954,26 @@ def estimate_completion():
                 break
             else:
                 raise
+
+        # Get the file system type from the given path
+        fstype = get_filesystem_type(pathname)
+        fstype = fstype[0]
+        print('File System:' + fstype)
```
az0, line 961: Please replace `print()` function with `logging.debug()`

bleachbit/FileUtilities.py
```diff
@@ -80,6 +80,36 @@ def open_files_linux():
     return glob.iglob("/proc/*/fd/*")
 
 
+def get_filesystem_type(path):
+    """
+    * Get file system type from the given path
```
az0, line 85: Please document the return value. What are the first and second values in the tuple? 

OR: It seems it only ever uses the first value, so maybe it doesn't need to return a tuple?

bleachbit/FileUtilities.py
```diff
@@ -924,12 +954,26 @@ def estimate_completion():
...
         # Write large blocks to quickly fill the disk.
         blanks = b'\0' * 65536
+        writtensize = 0
+        
         while True:
             try:
-                f.write(blanks)
+                if fstype != 'vfat':
+                    f.write(blanks)
+                elif writtensize < 4 * 1024 * 1024 * 1024 - 65536:
```
az0, line 972: Please add a code comment to explain the significance of this magic number.

az0 (changes requested):
This is much better than the last PR. Please see my four comments.

bleachbit/FileUtilities.py
Marvin0401: It is good idea. Let me modify it.

bleachbit/FileUtilities.py
Marvin0401: Ok. I will replace it.

bleachbit/FileUtilities.py
Marvin0401: The first values is file system type and the second value is a device name like "C://". Currently we are only using the first value, but I thought driver name could be needed in the future so I added it in return value. I will add the document for return value.

bleachbit/FileUtilities.py
Marvin0401: In the ubuntu system, there should be a error if the size of file is 4GB, no error if the size of file is less than 4GB. When we wipe the path, since we are writing 65536 bytes at a time if the size of space left is large than 4GB, the temp file size would be 4GB in the FAT32 system. In this case, the system always give us EFBIG error. I thought we need to set the maximum size of temp file as 4GB-65536 bytes.

Marvin0401: I will add the comment for this magic number.

az0:
I tested the fix for issue #1051 issue with AppVeyor build 1828 (with this commit) on Windows 10 against an 8GB FAT32 partition. It correctly created two temporary files. Here's the log. It looked good.

```
clean_operation('_gui'), options = '[]'
Wiping free disk space erases remnants of files that were deleted without shredding. It does not free up space.Wiping path: E:\
Creating new, temporary file for wiping free space.
Creating new, temporary file for wiping free space.
Wrote 2 files and 8,565,817,344 bytes in 53 seconds at 161.59 MB/s
Overwrite free disk space E:\ 0 

Disk space recovered: 0
Files deleted: 0
Special operations: 1
elapsed time: 53 seconds

```

Marvin0401:
Thank you.

PR #1112, opened 2021-02-03
Update thunderbird.xml

pas-calc:
the process is called `thunderbird`  (without bin) according to `ps -aux`


az0:
Yes, I verified on LinuxMint 19.3, CentOS 7, Ubuntu 16.04, and Ubuntu 20.10

PR #1128, opened 2021-02-21
new cleaner : pix ( history )

pas-calc:
deletes history of pix image browser

### DIFF
--- a/cleaners/pix.xml
+++ b/cleaners/pix.xml
@@ -1,0 +1,34 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+
+(c) Pascal Klein
+21/FEB/2021
+
+BleachBit CleanerML cleaner for:
+Pix (based on gThumb) photo browser
+
+    This program is free software: you can redistribute it and/or modify
+    it under the terms of the GNU General Public License as published by
+    the Free Software Foundation, either version 3 of the License, or
+    (at your option) any later version.
+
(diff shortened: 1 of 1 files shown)


az0:
Thank you for the submission, @pas-calc  :)

Would you please make two changes?

1. Add os="linux" to the cleaner element at line 24
2. Submit at https://github.com/bleachbit/cleanerml instead

pas-calc:
Now I have written a pull-request here : https://github.com/bleachbit/cleanerml/pull/289
So that means, new CleanerML files are not supposed to be written to the bleachbit/bleachbit repository? 

Please take into account, that I have asked about where to place a custom respectively new cleaner exactly, here : https://github.com/bleachbit/wishlist/issues/98 
So it might be worth explaining

az0:
> Now I have written a pull-request here : [bleachbit/cleanerml#289](https://github.com/bleachbit/cleanerml/pull/289)

Thank you

> So that means, new CleanerML files are not supposed to be written to the bleachbit/bleachbit repository?

Cleaners for the most popular applications go into bleachbit repository, while others go into the cleanerml  repository. Meanwhile, I am working on an online updater system for the cleanerml repository, so it will be more accessible
 
> Please take into account, that I have asked about where to place a custom respectively new cleaner exactly, here : [bleachbit/wishlist#98](https://github.com/bleachbit/wishlist/issues/98)
> So it might be worth explaining

I will take another look

PR #1139, opened 2021-03-23
Fix Windows explorer context menu command and add a test for it.

rados:
Please note that I moved several lines up the initializing of self._auto_exit in GUI.py->Bleachbit->__init__ method and I am not completely sure that it is safe.

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -122,6 +122,10 @@ class Bleachbit(Gtk.Application):
         Gtk.Application.__init__(
             self, application_id='org.gnome.Bleachbit', flags=Gio.ApplicationFlags.FLAGS_NONE)
         GObject.threads_init()
+        
+        if auto_exit:
+            # This is used for automated testing of whether the GUI can start.
+            self._auto_exit = True        
 
         if shred_paths:
             self._shred_paths = shred_paths
@@ -139,9 +143,6 @@ class Bleachbit(Gtk.Application):
             except ImportError:
                 logger.exception(
                     _("Error loading the SQLite module: the antivirus software may be blocking it."))
-        if auto_exit:
-            # This is used for automated testing of whether the GUI can start.
-            self._auto_exit = True
 
     def build_app_menu(self):
         """Build the application menu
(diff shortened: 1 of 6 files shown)


abitrolly:
Why did you move these lines?

rados:
> Why did you move these lines?

Because I wanted to use _auto_exit member variable as a flag if the application was started externally from Windows Explorer context menu. And if we initialize _auto_exit at the end of the constructor then it remains uninitilized because the constructor returns after "if shred_paths:". So in this case I cannot use it in do_activate method where the window is created. But this is an easy fix if these lines should be brought back at the end of the constructor...

abitrolly:
Does it make sense to introduce dedicated flag for this case?

rados:
> Does it make sense to introduce dedicated flag for this case?

I thought about it and decided to try to use existing one, if possible, in order to not complicate the code.

rados:
And while on the  subject, I am not sure that I understand the comment related with Bleachbit._auto_exit flag:
        if auto_exit:
            # This is used for automated testing of whether the GUI can start.

What kind of automated testing do we use?

abitrolly:
>  # This is used for automated testing of whether the GUI can start.

I don't understand it either. Maybe it means just testing if GUI can start. Clarifying this with a diagram would help understand the code better. Otherwise I need more time to study this code. Maybe @az0 can clarify.

az0:
> And while on the subject, I am not sure that I understand the comment related with Bleachbit._auto_exit flag:
> if auto_exit:
> # This is used for automated testing of whether the GUI can start.
> 
> What kind of automated testing do we use?

It is used by assert_execute_console() in `windows/setup_py2exe.py`. Some UPX optimizations  were too aggressive, causing the application not to start at all. See commit 90eaf2277

az0:
@rados 

I checked out your branch. On Linux, I ran `python3 bleachbit.py --gui --shred /tmp/asdf `. The GUI appears with the confirmation dialog, but the it crashes before I can respond. The console shows

```
Success
  File "/foo/bleachbit/bleachbit/Worker.py", line 260, in run
    for dummy in self.run_operations(self.operations):
  File "/foo/bleachbit/bleachbit/Worker.py", line 336, in run_operations
    name = backends[operation].get_name()
KeyError: '_gui'
```

On Windows, the behavior is the same.

az0:
On Windows 10, I installed build 1835, and the context menu works as expected (with one file).

az0:
I deleted my comment today about KeyError. This error was my fault. After testing properly, I confirm that this commit fixes the command line `--gui --shred`.

Issue #1072, opened 2020-12-21
Improve startup time with solving fontconfig cache issues under Windows


rados:
This is separate ticket about generating fontconfig cache causing app to be slow on startup. It is probably related with #1013.

Some guidelines:

1. Can the caching be avoided altogether because we need only certain fonts?
2. Are there cases where fc-cache fails to cache, so it tries to cache every time the application starts?
3. Can we give the user visual feedback that caching is in progress?
4. Can we run fc-cache from the installer without causing problems for special cases such as multi-user systems and silent installation mode?

az0:
Fixed

rados:
Reopen the issue because of this omission in the portable application.
https://github.com/bleachbit/bleachbit/pull/1144#issuecomment-813365114

az0:
@rados  Can this be closed because of 133f2ab7307b7ff9547d51a012dbd002e841db0c ?

rados:
> @rados Can this be closed because of [133f2ab](https://github.com/bleachbit/bleachbit/commit/133f2ab7307b7ff9547d51a012dbd002e841db0c) ?

Yes, exactly.

PR #1140, opened 2021-03-24
Fix #1013 #1072

### LINKED ISSUE #1013: Bleachbit 4 doesn't start in Windows 10 (slow start without feedback to user)
I recently tested Bleachbit 4 in Windows 10, but it didn't start at all. I believe I had the latest version; I installed it within the past 48 hours, unless you've updated Bleachbit within that time frame. In case it becomes important, I was running Windows 10 version 1909.

I do see a bleachbit.exe process in Task Manager, but no Bleachbit window opens. Running bleachbit_console.exe just opens a blank black window that does nothing. Bleachbit 2.0 ran just fine on the same installation of Windows, but I'd rather not have to go back to that.

rados:
Now we install only two fonts during nsis process and building the font cache on the first run is fast.

### DIFF
--- a/windows/bleachbit.nsi
+++ b/windows/bleachbit.nsi
@@ -226,7 +226,6 @@ VIFileVersion ${File_VERSION}
 ; MUI_UNPAGE_DIRECTORY not needed, ATM.
 ; !insertmacro MUI_UNPAGE_DIRECTORY
 !insertmacro MUI_UNPAGE_COMPONENTS
-UninstallText $(BLEACHBIT_UNINSTALL_TEXT)
 !insertmacro MUI_UNPAGE_INSTFILES
 !define MUI_UNFINISHPAGE_NOAUTOCLOSE
 !insertmacro MUI_UNPAGE_FINISH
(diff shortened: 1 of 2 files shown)


windows/bleachbit.nsi
```diff
@@ -226,7 +226,6 @@ VIFileVersion ${File_VERSION}
 ; MUI_UNPAGE_DIRECTORY not needed, ATM.
 ; !insertmacro MUI_UNPAGE_DIRECTORY
 !insertmacro MUI_UNPAGE_COMPONENTS
-UninstallText $(BLEACHBIT_UNINSTALL_TEXT)
```
abitrolly, line 229: Why this line is removed?

windows/bleachbit.nsi
```diff
@@ -309,15 +308,10 @@ Section Core (Required)
 
     SetOutPath $INSTDIR
     File /r /x "locale" "..\dist\*.*"
-    File "..\COPYING"
-
-    SetOutPath "$INSTDIR\share\"
-    File "..\bleachbit.png"
```
abitrolly, line 315: The same question. The COPYING files should not be removed from distribution.

windows/bleachbit.nsi
rados: This line is duplicated in the installer on line #427.

windows/bleachbit.nsi
rados: Because this is handled from the setup python file. Sorry, probably I had to split this commit and explain better...

PR #1144, opened 2021-04-02
Fix install Windows fonts for local user.

rados:
Just noticed that when we install for the local user the installer didn't copy the fonts, so I moved the lines responsible for this to a place where the fonts are copied for the local user.

### DIFF
--- a/windows/bleachbit.nsi
+++ b/windows/bleachbit.nsi
@@ -308,6 +308,9 @@ Section Core (Required)
 
     SetOutPath $INSTDIR
     File /r /x "locale" "..\dist\*.*"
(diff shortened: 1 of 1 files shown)


az0:
cross reference: this is a follow-up to https://github.com/bleachbit/bleachbit/pull/1140

XhmikosR:
Not sure where this broke but the current nightly portable doesn't have any fonts at all:

image

az0:
@XhmikosR  I think this is fixed in the [latest releases](https://ci.bleachbit.org/). If you still have an issue, please open a new issue ticket.

PR #1145, opened 2021-04-03
Fix #1040 by appending a subfolder (BleachBit) at the end of the cust…

### LINKED ISSUE #1040: Root installation directory is deleted during uninstallation
I have version 4.0.0.1628, Windows 10.
I want to uninstall but if I do, it starts deleting unrelated files and folders (from the same folder as the bleachbit install folder).
That's scary (I have to kill the process to stop the deletion, I already lost many files).

What should I do to uninstall Bleachbit?

rados:
…om installation path automatically.

### DIFF
--- a/windows/bleachbit.nsi
+++ b/windows/bleachbit.nsi
@@ -301,6 +301,14 @@ Function RefreshShellIcons
   System::Call 'shell32.dll::SHChangeNotify(i, i, i, i) v (${SHCNE_ASSOCCHANGED}, ${SHCNF_IDLIST}, 0, 0)'
 FunctionEnd
 
+Function .onVerifyInstDir
+  ; This callback belongs to MUI_PAGE_DIRECTORY and is called every time the user presses Browse button and selects install directory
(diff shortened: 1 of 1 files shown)


abitrolly:
An alternative is to avoid installation into non-empty dir. But I still think that that proper fix for #1040 is https://nsis.sourceforge.io/Uninstall_only_installed_files

If that dir contains any user level files, users probably don't want to see them cleaned.

rados:
> An alternative is to avoid installation into non-empty dir. But I still think that that proper fix for #1040 is https://nsis.sourceforge.io/Uninstall_only_installed_files
> 
> If that dir contains any user level files, users probably don't want to see them cleaned.

Yes, I checked quickly that option as well and decided to do it if it is explicitly requested. So, I can do it now. And offering an end node folder 'BleachBit' is not in conflict with it. It is good to have both - we can offer a folder and then we can uninstall only installed files. I just checked a couple of installers and they have both features in them.

PR #1148, opened 2021-04-17
Fix 1141: During unit testing use environment variable for setting options dir …

rados:
…and use TestAll.py to run all tests.

 Also I added mocks and they will be used when the environment variable is not set (while running individual tests for example).


abitrolly:
It looks like BleachBit doesn't allow to customize the location of options file. Why not to add this as a feature? Could be used for other purposes, not just for testing?

az0:
> It looks like BleachBit doesn't allow to customize the location of options file. Why not to add this as a feature? What are the other cases beyond testing where this can be useful?

It's an interesting idea, but the use cases are not clear and no users have asked for it, so I would rather focus now on more high-priority enhancements. If a power user needs this customization, it's still available via the environment variable. For now, I documented the new variable at https://docs.bleachbit.org/dev/testing.html

PR #1149, opened 2021-04-19
In portable application copy fonts before initializing GTK.

rados:


rados:
Closed because the branch name was incorrect. I will open a new one with correct branch name.

PR #1155, opened 2021-05-03
Tidy up Windows init

abitrolly:
There is no need to import Windows specific code in Linux, and `Windows` module is already conditionally imported below, 

I also removed `return` after `shred_files` that seems suspicious. With this `return`, the code reads "Do not clean up some nonce files if GUI started with `shred_paths`". I am not sure that `nonce` files are not created when this argument is used. Unless `shred_paths` is some special mode. In which case it would be nice to have its behavior documented in code comments.

@az0 PTAL.


az0:
I confirmed in recent build (1883) that the nonces are not deleted when shredding from context menu, so good catch. Thank you

PR #1167, opened 2021-05-29
Update sv for org.bleachbit.BleachBit.desktop

reportxx:


az0:
Thanks for the update. In the future please make this change in[ Launchpad](https://www.bleachbit.org/contribute/translate) because there the translations in the .desktop file come from LP.

PR #1176, opened 2021-06-14
Add option to not add desktop shortcut during command line installation.

rados:


windows/bleachbit.nsi
```diff
@@ -357,9 +357,16 @@ SectionGroup /e Shortcuts
     SectionEnd
 
     Section "Desktop" SectionDesktop
+        ${GetParameters} $R0
+        ${GetOptions} $R0 "/S /NoDesktopShortcut" $R0
```
az0, line 361: @rados 
Is there an easy way to make this (1) not case sensitive and (2) flexible to allow other orders such as `/S /AllUsers /NoDesktopShortcut`?

Issue #982, opened 2020-08-11
BleachBit adds wildcards to winapp2.ini file masks where there should be none


MoscaDotTo:
https://github.com/MoscaDotTo/Winapp2/issues/728#issuecomment-671223303

Quoted below

I have now checked it with BleachBit 4.0.0 and was able to reproduce the problem. However, the problem is not caused by the Winapp2.ini, but by a dangerous bug in BleachBit: BleachBit seems to put a wildcard before each search pattern!

Example: FileKey2=%LocalAppData%\Microsoft\Edge\User Data\Default|LOG;LOG.old|RECURSE

This FileKey should only delete files that are named exactly "LOG" or "LOG.old". BleachBit deletes however "*LOG" and "*LOG.old":
```
Delete 0 C:\Users\xxx\AppData\Local\Microsoft\Edge\User Data\Default\Local Extension Settings\iikmkjmpaadaobahmlepeloendndfphd\BleachBit.log
Delete 0 C:\Users\xxx\AppData\Local\Microsoft\Edge\User Data\Default\Local Extension Settings\iikmkjmpaadaobahmlepeloendndfphd\BleachBitlog
Delete 0 C:\Users\xxx\AppData\Local\Microsoft\Edge\User Data\Default\Local Extension Settings\iikmkjmpaadaobahmlepeloendndfphd\BleachBit.log.old
Delete 0 C:\Users\xxx\AppData\Local\Microsoft\Edge\User Data\Default\Local Extension Settings\iikmkjmpaadaobahmlepeloendndfphd\BleachBitlog.old
```
This is a very dangerous behavior for a cleanup tool and definitely a bug!

az0:
OK, I'd like to start by creating a unit test that reproduces this and to avoid regressions.

rados:
@az0 
I've just reproduced and started tracing the issue. I could finish it.

az0:
@MoscaDotTo 

Rados coded a fix for this, and would you please test [BleachBit build 1950 or later](https://ci.bleachbit.org/)?

APMichael:
I have repeated the test with version 4.4.1.1952. BleachBit now only deletes the files that are specified in the FileKey. The unwanted wildcard before each search pattern is now fixed. Thanks a lot!

az0:
> I have repeated the test with version 4.4.1.1952. BleachBit now only deletes the files that are specified in the FileKey. The unwanted wildcard before each search pattern is now fixed. Thanks a lot!

[...]

PR #1177, opened 2021-06-17
Fix regex generation to prevent accidental deletion of files with ove…

rados:
…rlapping names.


az0:
@rados this is for https://github.com/bleachbit/bleachbit/issues/982 ?

rados:
> @rados this is for #982 ?

Yes, exactly.

PR #1195, opened 2021-06-19
Create the standard toolbar

### LINKED ISSUE #809: Title bar with gtk3-nocsd
GTK 3 removed the title bar, but with [gtk3-nocsd](https://github.com/PCMan/gtk3-nocsd), we can re-add it.

But this does not work with Bleachbit 3.0.1 on Debian testing with gtk 3.24.13.

luigifab:
This is a solution for #809 for classic desktop environments (Mate and Xfce).

Before:
before

After:
after

Only tested with Debian Testing + Mate 1.24/1.26 + gtk3-nocsd.

I also added an item in Application menu, `Quit` with `Control+Q`.
I didn't search why the `Quit` menu isn't translated by GTK.

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -126,7 +126,7 @@ class Bleachbit(Gtk.Application):
         if auto_exit:
             # This is used for automated testing of whether the GUI can start.
             # It is called from assert_execute_console() in windows/setup_py2exe.py
-            self._auto_exit = True        
+            self._auto_exit = True
 
         if shred_paths:
             self._shred_paths = shred_paths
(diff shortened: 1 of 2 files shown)


az0:
@luigifab 
Thank you. I will check it out.

luigifab:
[gtk3-classic](https://github.com/lah7/gtk3-classic) can replace [gtk3-nocsd,](https://github.com/PCMan/gtk3-nocsd) and the initial "problem" is fixed.

naxl (approved):

luigifab:
Also, an id on the **window** node can be useful, for example `#BleachbitMainWindow`.
I would like to fix the position & style of the scrollbar with my theme.

image

luigifab:
Because of #1203, I have an extra width:

image

I can't remove it from my theme, and I didn't find any way.
But maybe you can don't do the `set_margin_right` when `GTK_OVERLAY_SCROLLING=0`?

PR #1208, opened 2021-07-02
Remove excessive outer widget.

rados:


Martii:
Works well on Focal with KDE. :smile: Thanks!

PR #1218, opened 2021-07-11
Use x is None rather than x == None

monosans:
Comparing is ~50% faster this way.


az0:
@monosans 
Thank you. This looks good. By the way, this code is from https://github.com/jsvine/markovify/

abitrolly:
Maybe just update the code to the latest version?

@monosans what do you use for profiling?

monosans:
> @monosans what do you use for profiling?

timeit

PR #1219, opened 2021-07-11
Remove unnecessary else branch

monosans:


monosans:
#1220

abitrolly:
Interesting. Why BlechBit needs Markov chains?

az0:
> Interesting. Why BleachBit needs Markov chains?

The Markov chain system is a fast way (compared to deep neural networks) to generate [chaff](https://docs.bleachbit.org/doc/chaff.html).

PR #1223, opened 2021-07-12
Be more clear about Discord Cache Clean

unknowntrojan:
Inform user that clearing Discord Cache might cause unintended clearing of favorited/pinned GIFs.

### DIFF
--- a/cleaners/discord.xml
+++ b/cleaners/discord.xml
@@ -30,7 +30,7 @@
   </var>
   <option id="cache">
     <label>Cache</label>
-    <description>Delete the web cache, which reduces time to display revisited pages</description>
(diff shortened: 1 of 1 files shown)


az0:
@unknowntrojan 

Do you which particular file contains this information? Maybe it could be split out into a separate option

unknowntrojan:
I do not, and I don't have any experience with electron apps and/or chromium, so I couldn't tell you.

abitrolly:
The way to figure this out is to backup `~/.config/discord`, then to pin/favor another GIF and compare the contents.

unknowntrojan:
According to the DevTools, the GIFs seem to be stored in the browser's "Local Storage". I was unable to locate the plaintext data anywhere. 
AVzwvWAudr

PR #1226, opened 2021-07-13
Update chromium cleaners

ebaldridge:
The Brave cleaner is deleting site settings when the history box is checked. I have added a new option to clean them and moved the line from history to this new option.

### DIFF
--- a/cleaners/brave.xml
+++ b/cleaners/brave.xml
@@ -176,7 +176,7 @@
   </option>
   <option id="site_preferences">
    <label>Site preferences</label>
-   <description>Shield settings and site permissions</description>
+   <description>Settings for individual sites</description>
     <action command="json" search="file" path="$$profile$$/Preferences" address="profile/content_settings/exceptions"/>
   </option>
 </cleaner>
--- a/cleaners/brave.xml
+++ b/cleaners/brave.xml
@@ -175,8 +175,8 @@
     <action command="sqlite.vacuum" search="file" path="%LocalAppData%\Google\Chrome\User Data\Default\Plugin Data\Google Gears\permissions.db"/>
   </option>
   <option id="site_preferences">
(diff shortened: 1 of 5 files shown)


cleaners/brave.xml
```diff
@@ -175,4 +174,9 @@
     <action command="sqlite.vacuum" search="file" path="%LocalAppData%\Google\Chrome\User Data\Default\Plugin Data\Google Gears\localserver.db"/>
     <action command="sqlite.vacuum" search="file" path="%LocalAppData%\Google\Chrome\User Data\Default\Plugin Data\Google Gears\permissions.db"/>
   </option>
+  <option id="site_preferences">
+   <label>Site preferences</label>
```
az0, line 178: It's good that "Site preferences" is already translated

cleaners/brave.xml
```diff
@@ -175,4 +174,9 @@
...
+   <description>Shield settings and site permissions</description>
```
az0, line 179: @ebaldridge 

The string "Settings for individual sites" is already translated (as used in Firefox, Waterfox, and Palemoon). Do you think this would be clear enough?

az0:
The same file is cleaned for other Chromium browsers: Chromium, Google Chrome, Microsoft Edge, and Opera, but they don't use a system called Shield. Still, it's good to have consistency with the options across these browsers. 

@ebaldridge 
Would you be willing to change them too?

cleaners/brave.xml
ebaldridge: I thought it would be good to keep the same name as on Firefox.

ebaldridge:
I have added the setting to the mentioned chromium browsers too. This will allow site-specific permissions to be kept when history is cleaned.

ebaldridge:
#1236 is fixed by this.

PR #1232, opened 2021-07-23
Fix NSIS language string.

rados:


az0:
@rados Thanks for fixing my mistake!

PR #1241, opened 2021-08-01
Update langsecref_map

livingentity:
Updated langsecref_map (mostly browsers) because winapp2.ini now uses some of them.


livingentity:
Oops, seemingly I messed it all up and unintentionally added additional commits here which I wanted as a separated pull request.
How do I fix this?

PR #1248, opened 2021-08-15
Make REMOVESELF delete the top-folder even with a globbed path

livingentity:
REMOVESELF is supposed to delete both the content and the top-folder. However though it succeeds for the content it currently fails to delete the _top-folder if it's path has a glob in it_. This commit fixes that.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -366,12 +366,10 @@ class Winapp:
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
             search = 'file'
-            type_str = ''
             if dirname.find('*') > -1:
                 search = 'glob'
-                type_str = ' type="d"'
-            action_str = '<option command="delete" search="%s" path="%s"%s/>' % \
-                         (search, xml_escape(dirname), type_str)
+            action_str = '<option command="delete" search="%s" path="%s" type="d"/>' % \
+                         (search, xml_escape(dirname))
             yield Delete(parseString(action_str).childNodes[0])
 
     def handle_filekey(self, lid, ini_section, ini_option, excludekeys):
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -365,8 +365,11 @@ class Winapp:
                      (search, xml_escape(path), regex, excludekeysxml)
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
-            action_str = '<option command="delete" search="file" path="%s"/>' % \
-                         (xml_escape(dirname))
+            search = 'file'
+            if dirname.find('*') > -1:
+                search = 'glob'
+            action_str = '<option command="delete" search="%s" path="%s" type="d"/>' % \
+                         (search, xml_escape(dirname))
             yield Delete(parseString(action_str).childNodes[0])
 
     def handle_filekey(self, lid, ini_section, ini_option, excludekeys):
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -365,11 +365,8 @@ class Winapp:
                      (search, xml_escape(path), regex, excludekeysxml)
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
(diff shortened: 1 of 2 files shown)


livingentity:
I now tested it on _all_ entries with REMOVESELF in the current winapp2.ini and it works fine.

However I identified a new problem:
If a folder isn't empty (e.g. because you only cleaned *.log but there are also other files or subfolders) it doesn't get deleted.

What do we want in this case? Don't know how CCleaner handles this.

Apart from that this commit should be ready to merge I guess.

az0:
FYI I started writing some unit tests for this. 

It's the folder is not empty, then it should not be deleted. Let's assume this until we can confirm otherwise.

livingentity:
Thank you.
Which variant do we want, commit 1 or 2? The 2nd one adds an additional `type="d"` for `search="file"` which isn't really needed but it shortens the code unless this causes a performance issue.

az0:
This is a [new test for REMOVESELF](https://github.com/bleachbit/bleachbit/commit/dfb4e08d05b07ffd6edd6802e03254a4b8fe236a), but it passes without the new code in this pull request. Even if you can't code it, please describe a test that is missing that would fail without the new code.

livingentity:
For safety I downloaded 4.4.1.2009, tried it and it does not remove folders whenever a glob is somewhere in the directory path, as expected.

Your test cases seem to be correct and should fail. I cannot easily tell why they pass, I would first need to learn how this works exactly.

livingentity:
Tried to fix [here](https://github.com/bleachbit/bleachbit/pull/1252).

az0:
> Tried to fix [here](https://github.com/bleachbit/bleachbit/pull/1252).

Thanks. Hmm, they are still passing.

For troubleshooting, can you suggest a section from Winapp2.ini to manually test?

az0:
Test did fail https://ci.appveyor.com/project/az0/bleachbit/builds/40469743#L2281

```
FAIL: test_removeself (tests.TestWinapp.WinappTestCase)
Test for the removeself option
  File "C:\projects\bleachbit\tests\TestWinapp.py", line 491, in test_removeself
    self.assertNotExists(fn, filekey)
  File "C:\projects\bleachbit\tests\common.py", line 122, in assertNotExists
    'The file %s should not exist, but it does. %s' % (path, msg))
AssertionError: The file C:\Users\appveyor\AppData\Local\Temp\1\WinappTestCaseqjr44qp8\dir_c\subdir\foo.log should not exist, but it does. FileKey1=C:\Users\appveyor\AppData\Local\Temp\1\WinappTestCaseqjr44qp8\dir_a\sub*|*.*|REMOVESELF
Ran 213 tests in 119.046s
FAILED (failures=1, skipped=34)
```

I will take a closer look

livingentity:
Yes, but that's just me being stupid. It shouldn't fail because of a file but because the top folder isn't deleted.

livingentity:
@rados
Just to rule out an isolated issue on my system, can you reproduce this (directly in BleachBit, I mean)? Please just take one entry of the winapp2.ini with REMOVESELF and a glob in the path, e.g. `%LocalAppData%\Google\Chrome*\Application\SetupMetrics|*.*|REMOVESELF` and check if the top-folder, i.e. SetupMetrics, gets deleted.

rados:
> @rados
> Just to rule out an isolated issue on my system, can you reproduce this (directly in BleachBit, I mean)? Please just take one entry of the winapp2.ini with REMOVESELF and a glob in the path, e.g. `%LocalAppData%\Google\Chrome*\Application\SetupMetrics|*.*|REMOVESELF` and check if the top-folder, i.e. SetupMetrics, gets deleted.

Yes it gets deleted.

livingentity:
Thanks for testing. To double check: Even without my commit it's deleted?

rados:
No, before the fix those with '*' in the path are not deleted.

livingentity:
Awesome. I already thought I lost my mind because andrew's tests suggest otherwise. Have a virtual beer. Cheers!

Issue #1247, opened 2021-08-12
Exclusion in winapp2.ini not honoured


livingentity:
**Describe the bug**
An excluded file in winapp2.ini is still deleted.

**To Reproduce**
1. Look at Firefox/Mozilla, Updates. It has `FileKey1=%CommonAppData%\Mozilla\updates|*.*|RECURSE` but also `ExcludeKey1=FILE|%CommonAppData%\Mozilla\updates\*\|update-config.json`.
2. The file gets deleted when cleaning. 

The error happens every time.

**Expected behavior**
The file should not be deleted.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - BleachBit version 4.4.1

rados:
@az0 
I can trace this one.

livingentity:
Fixed in PR [#1249](https://github.com/bleachbit/bleachbit/pull/1249).

az0:
Normally we wait to close the issue until the PR is merged, so I will reopen it for now. 

Thank you for working on this, livingentity. 

Now the change needs some review. Ideally you would add a new unit test (in `tests/TestWinapp.py`) such that it fails without the PR and passes with the new code. You can run the tests for this module by running `python3 -m unittest tests.TestWinapp` from the project's root directory. The test makes it easier to review.

livingentity:
I'm _not_ a programmer. I don't know about python or github either.
I just sat there, searched the project for 'ExcludeKey', found the right place and went through it line by line while looking up things in a python manual until I found the bug. Then I iteratively tried to find a fix which caused all this mess of a commit.

I would much prefer that someone with programming knowledge takes it from here.
(The actual issue is just that the very last backslash in `ExcludeKey1=FILE|%CommonAppData%\Mozilla\updates\*\|update-config.json`  is optional, so both cases, with or without it, need to be taken into account.)

PS: I reopen it because az0 wanted to do it but actually didn't.

rados:
@livingentity 
For non programmer you are pretty good. :)

livingentity:
Thanks again rados, I'm trying.

az0:
> I'm not a programmer. I don't know about python or github either.

Python uses duck typing: "If it walks like a duck, and it quacks like a duck, then it must be a duck.”  By this logic, if someone writes programs, then he is a programmer. 

> For non programmer you are pretty good. :)

I agree!

PR #1249, opened 2021-08-15
Fix for 'Exclusion in winapp2.ini not honoured' (issue #1247)

livingentity:


livingentity:
Only did some minor testing on the original issue [#1247](https://github.com/bleachbit/bleachbit/issues/1247) and a few other cases.

livingentity:
Now that I think of it there's probably a much easier solution:
Just replace `parts = excludekey.split('|')` by `parts = excludekey.split(r'\\*\|')`.

livingentity:
Tested with _all_ exclusion entries in the current winapp2.ini and it's fine.

However there's a different problem that file paths with square brackets don't work with regex because fnmatch.translate doesn't escape them.

But this can be merged, I guess.

rados:
@livingentity 
Nice work, thank you. 

@az0 
Seems OK to me.

PR #1250, opened 2021-08-18
Support square brackets in ExcludeKey path

livingentity:
Not sure it's worth it. Only few paths in winapp2.ini include square brackets but they do exist.

Example: `ExcludeKey1=FILE|%LocalAppData%\Packages\Microsoft.MicrosoftOfficeHub_*\AC\Microsoft\Internet Explorer\DOMStore\*\|www.office[1].xml`

Note for testing: This needs #1249 otherwise exclusions won't work at all anyhow.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -366,12 +366,10 @@ class Winapp:
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
             search = 'file'
-            type_str = ''
             if dirname.find('*') > -1:
                 search = 'glob'
-                type_str = ' type="d"'
-            action_str = '<option command="delete" search="%s" path="%s"%s/>' % \
-                         (search, xml_escape(dirname), type_str)
+            action_str = '<option command="delete" search="%s" path="%s" type="d"/>' % \
+                         (search, xml_escape(dirname))
             yield Delete(parseString(action_str).childNodes[0])
 
     def handle_filekey(self, lid, ini_section, ini_option, excludekeys):
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -133,14 +133,10 @@ def special_detect(code):
     return False
 
 
-def escape(matchobj):
-    return '[%s]' % matchobj.group(0)
-
-
 def fnmatch_translate(pattern):
     """Same as the original without the end and escaping square brackets"""
     import fnmatch
-    ret = re.sub('[][]', escape, pattern)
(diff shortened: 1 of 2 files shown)


rados:
@livingentity 
Makes sense. I would put the escape function in fnmatch_translate as a lambda function and would add some tests (in test_excludekey) to illustrate the problem and the solution...

livingentity:
Thanks for your suggestions. :)

livingentity:
Tests added.

Note to myself:
This allows square brackets in any part of the ExcludeKey. However in the FileKey square brackets might still not be escaped, e.g. because there's a glob and deletion fails. There's currently no entry for the latter case in the current winapp2.ini.

bleachbit/Winapp.py
```diff
@@ -134,9 +134,10 @@ def special_detect(code):
 
 
 def fnmatch_translate(pattern):
-    """Same as the original without the end"""
+    """Same as the original without the end and escaping square brackets"""
     import fnmatch
-    ret = fnmatch.translate(pattern)
+    ret = re.sub('[][]', lambda m: '[%s]' % m.group(0), pattern)
```
rados, line 139: @livingentity 
You can take a look at the format method of the string class in Python and use it in the lambda function instead of the old style with %.

tests/TestWinapp.py
```diff
@@ -429,6 +429,46 @@ def test_excludekey(self):
 
             # cleanup
             shutil.rmtree(dirname, True)
+            
+        # indices for tests when the ExcludeKey includes square brackets
```
az0, line 433: @livingentity 

The test_excludekey() function is long, and we want to avoid long functions. How about splitting these into their own test function like test_excludeke_square_bracket ? Another benefit is each function can run, even when the other function fails.

bleachbit/Winapp.py
livingentity: OK, but does that make any difference?

tests/TestWinapp.py
livingentity: Done.

rados (approved):

tests/TestWinapp.py
```diff
@@ -429,6 +429,46 @@ def test_excludekey(self):
 
             # cleanup
             shutil.rmtree(dirname, True)
+
```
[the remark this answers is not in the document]
rados: The idea with the new function is better anyway.

rados:
All tests pass locally (including the failed test here).

PR #1251, opened 2021-08-19
Actually delete IE webcache (e.g. dom cookies)

livingentity:
Some files in the IE webcache (including the most important one for history, cookies, etc.) were never actually deleted because they are always in use by Wininet\CacheTask. This commit fixes that.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -366,12 +366,10 @@ class Winapp:
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
             search = 'file'
-            type_str = ''
             if dirname.find('*') > -1:
                 search = 'glob'
-                type_str = ' type="d"'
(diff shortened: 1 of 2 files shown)


livingentity:
Winapp2.ini has many more entries. Do we want them, too?
Also I could make the system32 entries work for both 32-bit and 64- bit OS.

I'll have a look but not now.

livingentity:
Thinking of closing this for now because:
1. This cleaner is outdated anyhow and I don't want to fix it right now.
2. Even though stopping the CacheTask allows for deleting the WebCache file, it's immediately replaced with a new one and I don't know if it's a clean one or a full backup.
3. No one (else) seems to care about 64-bit cleaning but using BleachBit with Winapp2 will support it soon, i.e. Internet Explorer, too.

It's a little bit sad though that BleachBit cannot delete system cookies properly.
Any comments?

PR #1258, opened 2021-08-23
Expand Winapp REMOVESELF tests (fixed)

livingentity:


livingentity:
I see that existence is checked by calling `os.stat`. Using `os.stat(r'C:\Windows')` fails here, `os.stat('C:\\Windows')` works. Can someone please advice on the right handling of backslashes.

abitrolly:
@livingentity I would try `os.path.join('C:\', 'Windows')` https://docs.python.org/3/library/os.path.html#os.path.join

az0:
LGTM and AppVeyor could not test these because there was a merge conflict. This happens when two commits change the same lines of code, and a way to avoid it is this

1. Switch to the master branch
2. Update the master branch
3. Create a new branch from the master branch
4. Apply your changes there

In this case, I used the GitHub "resolve conflicts" button to manually fix it. Now I'll wait for AppVeyor

livingentity:
Thanks for resolving the conflict. Finally it fails the way it should.

PR #1259, opened 2021-08-26
Winapp2: Automatically translate System32 to Sysnative on 64-bit OS

livingentity:
Just a concept for now.

edit: Doesn't work at all, lol.

### DIFF
--- a/tests/TestWinapp.py
+++ b/tests/TestWinapp.py
@@ -435,6 +435,60 @@ class WinappTestCase(common.BleachbitTestCase):
 
             # cleanup
             shutil.rmtree(dirname, True)
+            
+    @common.skipUnlessWindows
+    def test_removeself(self):
+        """Test for the removeself option"""
+
+        # indexes for test
+        # position 0: FileKey statement
+        # position 1: whether the file `dir_c\subdir\foo.log` should exist after operation is complete
+        tests = (
+            # Refer to directory directly (i.e., without a glob).
+            (r'FileKey1=%s\dir_c|*.*|REMOVESELF' % self.tempdir, False),
+            # Refer to file that exists. This is invalid, so nothing happens.
+            (r'FileKey1=%s\dir_c\submarine_sandwich.log|*.*|REMOVESELF' %
+             self.tempdir, True),
+            (r'FileKey1=%s\dir_c\submarine*|*.*|REMOVESELF' % self.tempdir, True),
+            # Refer to path that does not exist, so nothing happens.
+            (r'FileKey1=%s\dir_c\doesnotexist.log|*.*|REMOVESELF' %
+             self.tempdir, True),
+            # Refer by glob to both a file and directory (which both start with `sub`).
+            # This should affect only the directory.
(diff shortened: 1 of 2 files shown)


livingentity:
Weird, did I mess up the backslashes? I could us `os.path.join(testpath, pathname)` instead of `'{}\\{}'.format(testpath, pathname)` anyways, but here the result is identical.
Btw why do people use such constructs if there's a dedicated function for the job?

rados:
On my local machine the tests pass, both - before and after the fix. At some point in the past test_detect_file didn't pass but I didn't explore why because it passed on the remote machine.

tests/TestWinapp.py
```diff
@@ -65,6 +65,20 @@ def get_winapp2():
     return fname
 
 
+def wow64_helper(path1, path2, testpath):
+    """test_detect_file helper function"""
+    dir1 = os.listdir(os.path.expandvars(path1))
+    dir2 = os.listdir(os.path.expandvars(path2))
+    dir_unique = set(dir2) - set(dir1)
```
rados, line 72: I am not sure why testing only unique entries gives some great benefit, because it is okay to have same folder in 'Program Files' and in 'Program Files (x86)'. Also when wow64_helper is called for 'Program Files' it returns only folders, probably it would be good to have some files as well.

livingentity: 1st, I didn't wrote that code. I just put it in a new function to make it re-usable.

2nd, the whole point of this test is to find out in which folder (32-bit or 64-bit) we land after redirection.
So first you _directly_ access 32-bit and 64-bit files or folders and take some unique to the one you want to land in. Then you try to access the unique stuff using the _redirection_ and see if it succeeds.

rados: I just thought for example we can get simply os.listdir(os.path.expandvars('%ProgramW6432%')) for the 64 bit case then do the append and that's all.

livingentity: There's no use in testing anything that is (or could be) present in both folders except there's a performance gain. Do you suggest that instead using the code above testing some (a lot for Sysnative) useless entries would be much faster?

rados: When os.getenv('ProgramW6432') is True we know that the OS is 64 bit, so we know the exact folder...

livingentity: Sorry for being bad at explaining.
> When os.getenv('ProgramW6432') is True we know that the OS is 64 bit, so we know the exact folder...

No, we don't know the exact folder, that's the whole point of this test. Redirection is happening outside of this test so we might end up somewhere else. Because we cannot trust the folder name we check it's content instead: If there are files which are unique to one folder and they are present we know we are in the right folder.

Checking _one unique_ file is sufficient for this test to work.
We already check _all unique_ files which is completely unnecessary but probably not a big deal.
Now you suggest checking _all_ files which is even more unnecessary and only works if there are unique files included which we would need to check first and therefore end up with a list of the unique files anyway.

This isn't about lessing the number of tested objects but getting the objects which are necessary for this test to work because non-unique objects don't.

PR #1264, opened 2021-08-31
Winapp compliance

livingentity:
Tried to make winapp behaviour more closely resemble the reference cleaner tool:

`*.*` means 'all files', with or without a '.'. So 'file_without_extension' gets deleted but 'folder1.0' doesn't match.

Only with RECURSE or REMOVESELF _empty_ folders are deleted afterwards.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -366,12 +366,10 @@ class Winapp:
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
             search = 'file'
-            type_str = ''
             if dirname.find('*') > -1:
                 search = 'glob'
-                type_str = ' type="d"'
-            action_str = '<option command="delete" search="%s" path="%s"%s/>' % \
-                         (search, xml_escape(dirname), type_str)
+            action_str = '<option command="delete" search="%s" path="%s" type="d"/>' % \
+                         (search, xml_escape(dirname))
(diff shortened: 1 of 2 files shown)


az0:
Maybe you already saw this error

https://ci.appveyor.com/project/az0/bleachbit/builds/40595076#L2276


```
======================================================================
FAIL: test_excludekey (tests.TestWinapp.WinappTestCase)
Test for ExcludeKey
  File "C:\projects\bleachbit\tests\TestWinapp.py", line 434, in test_excludekey
    test[3], r'%s\sub\deleteme.log' % dirname, msg)
  File "C:\projects\bleachbit\tests\common.py", line 132, in assertCondExists
    self.assertExists(path, msg)
  File "C:\projects\bleachbit\tests\common.py", line 117, in assertExists
    'The file %s should exist, but it does not. %s' % (path, msg))
AssertionError: The file C:\Users\appveyor\AppData\Local\Temp\1\bleachbit-test-winapp&34mw7f58\sub\deleteme.log should exist, but it does not. 
Test:
FileKey1=%(d)s|*eteme.*|RECURSE

```

livingentity:
Thanks but the above error happened after I tried to fix the original one.

The original one is in run_all -> validate_results. It expects the folder 'sub' to not exist but it does. The reason seems to be related to using ' type="f" '. Is the validator unable to handle it? (All tests for type only use ' search="file" ', globs are missing.)

PR #1268, opened 2021-09-05
Fix 1238 and 1227 (context menu) with adding new cmd parameter.

rados:


windows/NsisInclude/FilesToUninstall.nsh
```diff
@@ -323,7 +323,6 @@ Delete /REBOOTOK "$INSTDIR\libwebp-5.dll"
 Delete /REBOOTOK "$INSTDIR\libwinpthread-1.dll"
 Delete /REBOOTOK "$INSTDIR\libxmlxpat.dll"
 Delete /REBOOTOK "$INSTDIR\libzzz.dll"
-Delete /REBOOTOK "$INSTDIR\mfc100u.dll"
```
az0, line 326: @rados 
Do you now why this was removed? If not, no need to investigate

bleachbit/Windows.py
```diff
@@ -338,6 +338,16 @@ def elevate_privileges(uac):
     return False
 
 
+def _add_command_line_parameters(parameters):
+    """
+    Add any command line parameters such as --debug-log.
+    """
+    if '--context-menu' in sys.argv:
+        return '{} {} "{}"'.format(parameters, ' '.join(sys.argv[1:-1]), sys.argv[-1])
```
az0, line 346: @rados 
Would you please write a brief note here as a comment to explain why it is necessary to format it another way?

rados: Sure, because the file path looses its quotes and if it contains spaces it becomes a problem.

windows/NsisInclude/FilesToUninstall.nsh
rados: I don't. Actually those files  - FilesToUninstall.nsh and FilesToInstall.nsh are generated every time when we run the build procedure and probably we don't need to keep them in the repository. Back then I put them in just for reference and as a safe option because the fix was a recent one. Now I don't see a reason to keep them any longer in the repository...

abitrolly: Sometimes it is good to track autogenerated files in the repo for troubleshooting generators.

PR #1270, opened 2021-09-12
Support getting the windows version on Windows 8 or later

livingentity:
I did some more `winapp2.ini` related tests and was surprised that `DetectOS` didn't worked.

The reason is `Windows.get_windows_version()` which used `win32api.GetVersionEx(0)` to give a version that is being emulated for compatibility with the process, e.g. giving 6.2 for Windows 10. Now it's 10.0 instead.

### DIFF
--- a/bleachbit/Winapp.py
+++ b/bleachbit/Winapp.py
@@ -366,12 +366,10 @@ class Winapp:
         yield Delete(parseString(action_str).childNodes[0])
         if removeself:
             search = 'file'
-            type_str = ''
             if dirname.find('*') > -1:
                 search = 'glob'
-                type_str = ' type="d"'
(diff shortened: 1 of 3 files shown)


livingentity:
What's the difference between get_windows_version and parse_windows_build? One seems to be redundant.

livingentity:
After merging #1271 get_windows_version will get obsolete so I prepared a commit to completely remove it. Should I already merge this here?

rados:
@az0
Probably that 'obsolete' was about get_windows_version which is removed in this PR (not in 1271). In master we still use GetVersionEx which gives 6.2 for Windows 10 and messes things up. This PR replaces GetVersionEx with platform.win32_ver.

livingentity:
Sorry, no. This software is broken beyond repair. One 5 year old  bug more or less doesn't matter.

PR #1271, opened 2021-09-13
Remove some WinXP relicts

livingentity:


abitrolly:
For the better understanding what is going on on Windows, it would be nice to a get a table of output from these functions on different OS versions.

livingentity:
That might be nice for better understanding but BleachBit officially only supports Windows 8 and later so I'm not going to test what this might break in Windows XP.

tests/TestFileUtilities.py
```diff
@@ -338,13 +338,6 @@ def symlink_helper(link_fn):
             self.assertNotExists(linkname)
             self.assertNotLExists(linkname)
 
-        windows_vista_or_newer = False
-        if 'nt' == os.name:
```
az0, line 342: @livingentity 
It looks like we need to keep the line `if 'nt' == os.name:` so the kernel32.dll part does not run on posix systems. Otherwise, changes look good

livingentity: Thanks, done.

PR #1272, opened 2021-09-13
Support winapp2 64-bit regkey detection

livingentity:


livingentity:
Works but needs some rethinking.

PR #1319, opened 2021-12-25
get_volume_bitmap: cast buf_size to int

mkhon:


bleachbit/WindowsWipe.py
```diff
@@ -648,7 +648,7 @@ def get_volume_bitmap(volume_handle, total_clusters):
     input_struct = struct.pack('q', 0)
 
     # Figure out the buffer size. Add small fudge factor to ensure success.
-    buf_size = (total_clusters / 8) + 16 + 64
+    buf_size = int(total_clusters / 8) + 16 + 64
```
abitrolly, line 651: Should it be rounded up?

mkhon: There's some fixup being added already (+16+64)

Also, Python 2.7 (for which this code was initially written) implements / just as integer division (meaning, rounding down)

PR #1326, opened 2021-12-27
Add support GIMP 2.10 and detection if GIMP runs

Finoway:
The pull-request adds compatibility with GIMP version 2.10 and detection if GIMP runs.


cleaners/gimp_new.xml
```diff
@@ -22,6 +22,8 @@
 <cleaner id="gimp_new">
   <label>GIMP 2.10</label>
   <description>Graphics editor</description>
+  <running type="exe" os="windows">gimp-2.10.exe</running>
+  <running type="exe" os="linux">gimp-2.10</running>
```
abitrolly, line 26: That works when GIMP is started from GNOME. If could also be started from shell as `gimp` and then detection will fail.
```
|   \_ gimp
|       \_ /usr/lib64/gimp/2.0/plug-ins/script-fu/script-fu -gimp 10 9 -run 0
```

az0:
Thank you for the updates, @Finoway  . Please combine into one XML file.

cleaners/gimp_new.xml
Finoway: Hello,
I just tested on ubuntu as you can see on my screenshot and also on ArchLinux KDE by opening GIMP with the terminal
Capture d’écran de 2021-12-29 19-15-28

Finoway:
> Thank you for the updates, @Finoway . Please combine into one XML file.

Hello, thank for the answer I will try to do this in the night (I am French).

cleaners/gimp.xml
```diff
@@ -22,12 +22,35 @@
 <cleaner id="gimp">
   <label>GIMP</label>
   <description>Graphics editor</description>
+    <!-- GIMP 2.4 -->
```
az0, line 25: Comments are not needed because it's clear from the executable names

PR #1357, opened 2022-04-19
Update TestDeepScan.py

verdy-p:
safer exclusion of `*~` backup files starting by a file, and the last character before `~` may be also a digit (e.g. `index.php7~` or `README2~`

### DIFF
--- a/tests/TestDeepScan.py
+++ b/tests/TestDeepScan.py
@@ -75,7 +75,7 @@ class DeepScanTestCase(common.BleachbitTestCase):
         """Unit test for class DeepScan.  Preview real files."""
         path = os.path.expanduser('~')
         searches = {path: []}
-        for regex in ('^Makefile$', '~$', 'bak$', '^Thumbs.db$', '^Thumbs.db:encryptable$'):
+        for regex in ('^(Makefile|Thumbs(:encryptable)?\\.db|[^.].*[0-9A-Za-z]~|bak)$'):
             searches[path].append(Search(command='delete', regex=regex))
         ds = DeepScan(searches)
         for cmd in ds.scan():


abitrolly:
Test are broken. Are all commits pushed to this PR?

az0:
@verdy-p Yes, please address the build error as  @abitrolly mentioned

For convenience, here is the error from AppVeyor

```
======================================================================
ERROR: test_DeepScan (tests.TestDeepScan.DeepScanTestCase)
Unit test for class DeepScan.  Preview real files.
  File "C:\projects\bleachbit\tests\TestDeepScan.py", line 81, in test_DeepScan
    for cmd in ds.scan():
  File "C:\projects\bleachbit\bleachbit\DeepScan.py", line 105, in scan
    compiled_searches = [CompiledSearch(s) for s in searches]
  File "C:\projects\bleachbit\bleachbit\DeepScan.py", line 105, in <listcomp>
    compiled_searches = [CompiledSearch(s) for s in searches]
  File "C:\projects\bleachbit\bleachbit\DeepScan.py", line 67, in __init__
    self.regex = re_compile(search.regex)
  File "C:\projects\bleachbit\bleachbit\DeepScan.py", line 65, in re_compile
    return re.compile(regex, fs_scan_re_flags) if regex else None
  File "c:\python34\lib\re.py", line 223, in compile
    return _compile(pattern, flags)
  File "c:\python34\lib\re.py", line 294, in _compile
    p = sre_compile.compile(pattern, flags)
  File "c:\python34\lib\sre_compile.py", line 568, in compile
    p = sre_parse.parse(p, flags)
  File "c:\python34\lib\sre_parse.py", line 780, in parse
    p = _parse_sub(source, pattern, 0)
  File "c:\python34\lib\sre_parse.py", line 377, in _parse_sub
    itemsappend(_parse(source, state))
  File "c:\python34\lib\sre_parse.py", line 730, in _parse
    raise error("unbalanced parenthesis")
sre_constants.error: unbalanced parenthesis
```

Full log: https://ci.appveyor.com/project/az0/bleachbit/builds/43277935#L2245

abitrolly:
@verdy-p did you run the test you've modified yourself? I see at least one logical error in this modification.

verdy-p:
Yes I have run this in BleachBit itself. It works perfectly (and it is considerably faster too! in fact the "for" loop here is not necessary as it will loop only once, but it has been kept)

What can of logical error do you see? Here the regexp is anchored on both sides (start and end), and the two regexps  '~$', 'bak$' have been anchored too, by prefixing them by '[^.].*' (to exclude "hidden" filenames starting by a dot, that should not be "cleaned", then match remaining characters of the filename). This change however does not invalidate the regexp.

This is a one-line modification, only in the regexp string, and the regexp is perfectly valid, in Python (the parsing made by AppSurveyor is clearly wrong, all parentheses are correctly paired).

Note that AppSurveyor compiles the regexp in the source, but does not show which "flags" it uses to compile the constant string containing the regexp. Does AppSurveyor use the standard regexp engine of Python (the same engine as used by Bleachbit itself)?

az0:
@verdy-p 

Want to try with the new mkhon-python310 branch? It is a major update of the Python version

az0:
@verdy-p 
It's been a while, so I am checking in on how you'd like to continue.

abitrolly:
> What can of logical error do you see?

@verdy-p when you "for loop" over a string, you get separate characters.

```python
>>> for x in ('ddfsdfdsf'):
...     print(x)
...     
d
d
f
s
d
f
d
s
f
````

az0:
I re-tested the change on the latest code, and it fails. As abitrolly pointed out, it's looping over a string.

I do see a performance benefit of combining regexes, and I do see a CPU bottleneck when scanning a very large home directory on an SSD with warm cache.  In case of slower file system (e.g., HDD with cold cache), the CPU may not be the bottleneck.

For simply improving the test, there's alternatives. The test would work with one or two regexes, or over a controlled, small directory tree.

The real deep scan could be modified to replace regexes like *.bak with ends_with(), which I benchmarked as noticable faster.

When Python 3.14 without GIL becomes popular, it may be an option to use multithreading.

For today, I'm closing this PR because no response lately from verdy-p. I do apologize for my part in the delays of the conversation.

If you want to continue, feel free to re-open.

PR #1371, opened 2022-05-18
Add a KDE5 folder (RecentDocuments)

Martii:
* Affects at least Kubuntu 20.04 to 22.04

NOTES:
* I'm not real comfortable with obliterating ~/.cache all in one shot for KDE 5... there is a ~/.cache folder ... and has at least thunderbird *(distro installed)* which is [configured elsewhere in current thunderbird.xml](https://github.com/bleachbit/bleachbit/blob/ff06d09a89072dc064e97816ae81b43788610f66/cleaners/thunderbird.xml#L42) cleaner ... so skipping
* I don't see any kind of general .tmp in my user profile *(~)* ... so skipping
* This only clears the cached *(written)* list and **not** the in-memory *(Note these persist somewhere hidden atm between reboots/sessions)* "Places" *(22.04.x)* and "History" *(20.04.x)*

Historical refs:
* #130
* https://github.com/bleachbit/bleachbit/commit/78684e510c7334392b991f3a7efc5e435cf3cd3e
* https://github.com/bleachbit/bleachbit/commit/ae95d012f560a0fb970122ca741ad6fef114a815
* https://github.com/bleachbit/bleachbit/commit/aec8ac1b4b8ab10e430f5b8297ecd1c74df9d1a5

### DIFF
--- a/cleaners/kde.xml
+++ b/cleaners/kde.xml
@@ -43,5 +43,6 @@
     <description>Delete the list of recently used documents</description>
     <action command="delete" search="glob" path="~/.kde/share/apps/RecentDocuments/*.desktop"/>
     <action command="delete" search="glob" path="~/.kde4/share/apps/RecentDocuments/*.desktop"/>
+    <action command="delete" search="glob" path="~/.local/share/RecentDocuments/*.desktop"/>
   </option>
 </cleaner>


Martii:
Thanks for the consideration. :smile_cat:

Martii:
@az0 

Took a while but I reverse engineered a few things and this command line:

``` sh
qdbus org.kde.ActivityManager /ActivityManager/Resources/Scoring org.kde.ActivityManager.ResourcesScoring.DeleteStatsForResource :current :any "*"
```

... clears the in memory MRU *(Applications and Documents... quotes may be needed on all 3 trailing parameters in case something else ever goes there and the parameters must be present or qdbus will throw a false error message)* in KDE5. The default Places->History in Application Launcher does it a little different with the last parameter being `"/*"`, then another call to `"smb:*"`, then another call to `"/fish:*"`, and another call to `"/sftp:*"`, and a final call to `"/ftp:*"`. I surmised that it would take a global glob asterisk and it did clear everything out that I had previously opened including sftp and fish *(didn't know about this Dolphin protocol until now :smile_cat:  )*. 

Couple of quick questions...

1) Shall I make a new issue for this?
2) Does bleachbit have a way to call a command line *(exec call)*? If so I can possibly do another PR to "fill in the blank" part of this PR.
3) There's also one for Klipper *(The default KDE clipboard)*:

``` sh
qdbus org.kde.klipper /klipper org.kde.klipper.klipper.clearClipboardHistory
```

Please let me know especially item 2 and how you want to proceed?

Thanks mate.

PR #1398, opened 2022-09-25
Cleanup codebase

kenrap:


abitrolly:
-0.5 on doing code reformatting without functional changes. It makes `git blame` hard to navigate.

There is also some strange sentimental value that BleachBit code stands out specifically because it is not normalized, but I can't decide if it bug or feature with my perception.

kenrap:
Should I close this PR then?

I was already in process of converting several string concatenations and old-style formatting to using f-strings (which also improves string performance, not just readability) and was planning to work on refactoring the codebase. Sadly, if I'm going to get slapped in the face because of not following the sentimental value, then I won't bother. I'm sorry if that's the case.

kenrap:
Since I mentioned about it, I'll throw in what I already have.

The f-string work is not finished but it won't matter if I continue to get harsh feedback anyway.

abitrolly:
@kenrap let's @az0 make the decision. I am a master of harsh feedback, but not the one who manages this project, so don't pay much attention to it. If you can find an issue that you can fix, or functional code improvement, I will be interested to give you a proper review. So don't feel discouraged.

If you really want to see the code reformatted, I think that adding `black` to CI/CD pipeline, with the ability that bot does autocorrections if asked, would be an awesome addition that saves everyone's time. But the last word is still after Andrew.

bleachbit/FileUtilities.py
```diff
@@ -237,13 +237,12 @@ def write(parser, ini_file):
         removing a cast to str. This is needed to handle unicode chars.
         """
         if parser._defaults:
-            ini_file.write("[%s]\n" % "DEFAULT")
+            ini_file.write("[DEFAULT]\n")
```
kenrap, line 240: I'm just leaving a note here to prevent possible confusion.

The reason I left out the `f` in this statement was because it's formatting from a literal string. Thus it's better to inline it into the string it's formatting.

kenrap:
Just to let you folks know, I have tested these latest changes and it seems to work fine without any changes in functionality. 

Bleachbit feels like it removes through files faster, but I haven't benched it (nor do I know how to bench Python code) to prove it otherwise.

az0:
I haven't yet reviewed this in detail, but some initial feedback.

1. A concern with changing translated stings is that they will need translated again to >60 languages. Not all translators are active, so this could take a long time. This is more acceptable for logging messages, which are not often seen, but let's please avoid it

1. Smaller PRs are easier to review, so they are quicker to merge.

> Btw, what's the minimum supported Python version in Bleachbit?

Right now it's Python 3.4, but there is an issue to fix tis . See this issue https://github.com/bleachbit/bleachbit/issues/848 . 

> Bleachbit feels like it removes through files faster, but I haven't benched it (nor do I know how to bench Python code) to prove it otherwise.

Unless you are improving the disk I/O speed related to deleting files, you can benchmark it using the preview mode (i.e., not deleting any files). Perhaps it is not correct, but there are a few benchmarks for specific functions inside the test suite that can be run using `make tests`. Otherwise you can do a preview like this: `$ time ( python3 bleachbit.py -p --all-but-warning > /dev/null)` . The first time this is run, it will be slowest because of storage cache, but since you are not testing this part, ignore the first run and average it over a few runs (after the first run).

kenrap:
@az0 
> A concern with changing translated stings is that they will need translated again to >60 languages. Not all translators are active, so this could take a long time. This is more acceptable for logging messages, which are not often seen, but let's please avoid it for the typical UI messages, like the toolbars, dialogs, and preferences menu

Understood! I'll revert the string changes related to `_()` localization calls.

> Smaller PRs are easier to review, so they are quicker to merge.

I could split the work into different commits for each file and resubmit them as separate PRs, if that would make it easier for you?

> Right now it's Python 3.4, but there is an issue to fix tis . See this issue #848 .

Ooph! I'll need at least 3.6 support to use f-strings. I'll take a look into solving this issue, since I also would like to get 3.8 support at the minimum as stated before.

> Unless you are improving the disk I/O speed related to deleting files, you can benchmark it using the preview mode (i.e., not deleting any files). Perhaps it is not correct, but there are a few benchmarks for specific functions inside the test suite that can be run using make tests. Otherwise you can do a preview like this: $ time ( python3 bleachbit.py -p --all-but-warning > /dev/null) . The first time this is run, it will be slowest because of storage cache, but since you are not testing this part, ignore the first run and average it over a few runs (after the first run).

I appreciate the advice on the time testing. I've refactor a lot of parts of Cleaner.py that involved `Command.Delete` and `Command.Function` using generator expressions and `yield from` (to also help lower the level for loop nesting in some cases). I suppose since the code involves `yield` it might not have a performance improvement given the body of code is already in a generator function. When I get around to timing the changes, I will report back the results.

Also, thank you for reviewing! It's appreciated!

kenrap:
Removed code formatting commit, removed f-string usage in `_()` calls, and rebased from upstream master.

And the diff count got significantly trimmed down. Awesome!

kenrap:
@az0 I decided to remove all of the f-string usage with this latest push since I think it's best to get this merged sooner without needing #848 to be solved. Since the diff count got further trimmed down, this will make it easier for you to review.

kenrap:
Regarding benching the I/O performance, I wrote a simple bencher decorator for testing it. And there is practically no performance difference compared to the 4.4.2 release with my the latest push.

I'm aware of `timeit` but I read that it was slow. So I wrote this decorator instead in an untracked `my_bencher.py` file for future reuse:
```python
from datetime import datetime
import sys


def benches(reps):
    def decorator(func):
        def wrapper(*args):
            timings = []
            for n in range(1, reps + 1):
                start = datetime.now()
                func(*args)
                end = datetime.now()
                timing = end - start
                print(f'Execution #{n}: {timing.microseconds} microseconds', file=sys.stderr)
                timings.append(timing.microseconds)
            print(f"Average timing: {float(sum(timings))/reps} microseconds", file=sys.stderr)
        return wrapper
    return decorator
```
And decorated the appropriate function in `bleachbit/CLI.py` to work with using the `python3 bleachbit.py -p --all-but-warning > /dev/null` command mentioned earlier.
```python
@my_bencher.benches(5)
def args_to_operations_list(preset, all_but_warning):
...
```

kenrap:
I might as well remove the tuple conversions in my refactoring here. Yet another opportunity to make it easier for reviewing the changes by shrinking the diff.

Profiling Python code is a skill I really need to pick up. It's better than making _premature_ assumptions. 😉

For all I know, Python 3.11 or even 3.10 (or earlier) would make such conversions not necessary. That's why I need to learn to profile code.

kenrap:
@abitrolly So, with this latest push, will this PR's diff earn your "relatively clean and still git-blame navigable"  seal of approval? 😋

abitrolly:
@kenrap I will be +1 with Andrew on that.

> Smaller PRs are easier to review, so they are quicker to merge.

Even if each line is covered by test, it will still take a long time to review the logic in the whole file, and I need to find a job better sooner that later, so I am not ready to get into that. Realistically, it is over my budget. Take the time needed to review 4 lines in #1404 and then multiply by `max(diff_added, diff_removed)/4`, which is `111/4=27.75`

Code that fixes issues or even better - builds up on other PRs has a different priority, because it potentially improves my workflow. Just being honest about that.

abitrolly:
> "relatively clean and still git-blame navigable" seal of approval

With "relatively", yes. :D But that doesn't include "this code doesn't break anything" certificate.

kenrap:
Gotcha! I will be cognizant about the reviewer's time and effort.

While this kind of work does have a lower priority for review, I do think it can help the codebase in the longterm with improvements to its maintainability. I suppose that's all subjective in the end though.

Good luck on your job hunt! 👍

bleachbit/Action.py
```diff
@@ -71,7 +71,7 @@ def expand_multi_var(s, variables):
     for var_set in vars_product:
         ms = s  # modified version of input string
         for var_key, var_value in var_set.items():
-            sub = '$$%s$$' % var_key
+            sub = '{}{}{}'.format(var_format, var_key, var_format)
```
az0, line 74: This string formatting change is okay. It's longer, though. What is the benefit you see?

bleachbit/Cleaner.py
```diff
@@ -74,15 +75,15 @@ def add_action(self, option_id, action):
         """Register 'action' (instance of class Action) to be executed
         for ''option_id'.  The actions must implement list_files and
         other_cleanup()"""
-        self.actions += ((option_id, action), )
+        self.actions.append((option_id, action))
```
az0, line 79: This is fine

bleachbit/Cleaner.py
```diff
@@ -479,10 +478,14 @@ def gtk_purge_items():
                     '~/.recently-used.xbel',
                     '~/.local/share/recently-used.xbel*',
                     '~/snap/*/*/.local/share/recently-used.xbel']
-            for path1 in xbel_pathnames:
-                for path2 in glob.iglob(os.path.expanduser(path1)):
-                    if os.path.lexists(path2):
-                        yield Command.Shred(path2)
+            shred_paths = (
```
az0, line 481: This is tider without having to add more indentation. This syntax is uncommon, though, and I wonder whether other people would have a hard time reading it, so I am not sure this is worth it. What is it called?

bleachbit/Command.py
```diff
@@ -122,10 +122,10 @@ def __init__(self, path, func, label):
             raise AssertionError('Expected MethodType but got %s' % type(func))
 
     def __str__(self):
+        path = ''
```
az0, line 125: This could be confusing: path vs self.path

bleachbit/__init__.py
```diff
@@ -112,10 +112,14 @@
...
-    e3 = os.path.exists(os.path.join(bleachbit_exe_path, '../COPYING'))
-    portable_mode = all((e1, e2, e3))
+    paths = (
+        '../cleaners',
+        '../Makefile',
+        '../COPYING')
+    existing = (
+        os.path.exists(os.path.join(bleachbit_exe_path, path))
+        for path in paths)
+    portable_mode = all(existing)
```
az0, line 122: good

az0 (commented):
Some changes are easy to accept, but I am not sure about some rewrites. Please see my question about the syntax. 

If you want to split it up, I can accept some earlier, or we can continue working through this.

bleachbit/Action.py
kenrap: Thank you for taking time to review this.

The benefit here would be that if `$$` needed to be changed to some other token, changing the string value from the `var_format` constant variable is only needed. It's just a maintenance improvement.

To make it shorter, `var_format` could be renamed to `var_fmt` or something more appropriate. Since it's a constant variable, it probably needs to be all caps to notate that it's a constant as well.

bleachbit/Cleaner.py
kenrap: The syntax here is called a generator expression. It's a "lazy" version of the list comprehension syntax. I think list/generator expressions are relatively popular now-a-days, however I can reset this specific diff hunk with no problems! 👍

bleachbit/Command.py
kenrap: Good point. How about renaming `path` to perhaps `info` or something better?

kenrap:
@az0 
> If you want to split it up, I can accept some earlier, or we can continue working through this.

I'm good with continuing through this. I'm in no rush.

bleachbit/Cleaner.py
rados: @kenrap Probably more readable will be if we clean the nested loops with 'map' and 'filter' for example...

kenrap: @rados May I have you provide me an example?

On a technical note, I think Guido would be in disagreement there...

https://www.artima.com/weblogs/viewpost.jsp?thread=98196

rados: @kenrap Okay, I see, thank you. Then it could be made flat with two generator expressions. You can give it a try and see if it is more readable...

kenrap: @rados Hmm, with `glob.iglob` returning a glob iterator from each path in `xbel_pathnames`, the problem here is, even if I split it off, I would still need to do that "multi-line generator expression" in order to "flatten" out the glob iterator in the second generator expression. This is because the first generator expression would be a "generator of iterators."

```python
glob_iters = (glob.iglob(os.path.expanduser(path)) for path in xbel_pathnames)
shred_paths = (
    path
    for glob_iter in glob_iters
    for path in glob_iter
    if os.path.lexists(path))
```

I think the above approach would be less desirable?

kenrap: I had an idea of using the REPL to help with visualizing the problem and wanted to share this despite it not being necessary:
```python
>>> import glob, os
>>> paths = ["/var", "/foo", "/mnt"]
>>> glob_iters = [glob.iglob(os.path.expanduser(path)) for path in paths] # pretend this is a generator since they are exhaustible
>>> glob_iters
[<generator object _iglob at 0x7fcb6525b450>, <generator object _iglob at 0x7fcb6525b4c0>, <generator object _iglob at 0x7fcb6525b530>]
>>> flatten = [path for glob_iter in glob_iters for path in glob_iter]
>>> flatten
['/var', '/mnt']
```
Although it's interesting how using `glob.iglob` already filtered out the non-existent path without needing to use `os.path.lexists`.

abitrolly:
> And recently I have found other means of satistying that boredom now.

Like https://adventofcode.com/ or https://about.gitlab.com/community/hackathon/ ? )

> It really hurts my feelings when I've kept getting dismissed for this kind of work and my other ideas from @abitrolly. It makes me want to contibute less.

Yea. This part always hurts, so just get used to it, because in real job you may spend months polishing something that eventually will be discarded. It is not the code that you write now that matters, but the code you write in the future, so take it as an exercise how not to be attached to it.

Enjoy you time, and thanks for the @1404.

romelevans:
Hi Kenneth,

   I’m sorry for the lack of response.  I’m a Software Engineer/Consultant.  I was so busy with work I couldn’t respond before, but I’ll try to put something together for you as I can.

 

Thanks,

  Romel

az0:
@kenrap 

I understand how you feel as a contributor, as I have had difficulty in getting my contributions accepted in some projects. Let's work on some bite-size changes that are easier to accept.

BleachBit on Windows is now Python 3.10, and the oldest Linux distribution is Ubuntu 18.04 (bionic) which has Python 3.6, so f-strings can safely be used.  Interested in making this change?

bleachbit/Cleaner.py
az0: @kenrap  Resubmit?

bleachbit/Cleaner.py
```diff
@@ -202,38 +200,39 @@ def __init__(self):
         # reference: http://katana.oooninja.com/w/editions_of_openoffice.org
         if 'posix' == os.name:
             self.prefixes = ["~/.ooo-2.0", "~/.openoffice.org2",
-                             "~/.openoffice.org2.0", "~/.openoffice.org/3"]
-            self.prefixes += ["~/.ooo-dev3"]
+                             "~/.openoffice.org2.0", "~/.openoffice.org/3",
+                             "~/.ooo-dev3"]
```
az0, line 204: @kenrap  Resubmit?

bleachbit/Cleaner.py
```diff
@@ -74,15 +75,15 @@ def add_action(self, option_id, action):
...
+        self.actions.append((option_id, action))
 
     def add_option(self, option_id, name, description):
         """Register option (such as 'cache')"""
         self.options[option_id] = (name, description)
 
     def add_running(self, detection_type, pathname):
         """Add a way to detect this program is currently running"""
-        self.running += ((detection_type, pathname), )
+        self.running.append((detection_type, pathname))
```
az0, line 86: @kenrap  Resubmit?

bleachbit/Cleaner.py
```diff
@@ -156,12 +157,9 @@ def is_running(self):
...
             pathname = running[1]
-            if 'exe' == test and 'posix' == os.name:
-                if Unix.is_running(pathname):
-                    logger.debug("process '%s' is running", pathname)
-                    return True
-            elif 'exe' == test and 'nt' == os.name:
-                if Windows.is_process_running(pathname):
+            if 'exe' == test:
+                if ('posix' == os.name and Unix.is_running(pathname)) or \
+                   ('nt' == os.name and Windows.is_process_running(pathname)):
```
az0, line 162: @kenrap  Resubmit?

bleachbit/__init__.py
az0: @kenrap  Resubmit?

PR #1415, opened 2022-10-30
cli: Suppress `--context-menu` option from help output

abitrolly:
Reorders hidden flags towards the end and hides `--context-menu`.

I didn't spend much time looking into why `--context-menu` adds `--exit` and `--gui`, so just adding a comment here to make it easier to research later.


abitrolly:
@az0 thanks. That magically unblocks me to do other things. :D

Issue #1183, opened 2021-06-19
Make Chaff popup should be automatically closed upon opening another menu


banaagravrommel:
**Describe the bug**
Make chaff popup should be closed if another popup has been opened. 
I would like it if the "make chaff" popup will automatically closed if another popup has been opened like the "preference" popup appears because application should not allow multiple popup open while the other is not close

**To Reproduce**

1. Steps to reproduce the behavior:
2. Open the application
3. Click BleachBit Icon on the top right
4. Select Make Chaff
5. Click BleachBit Icon on the top right
6. Select Preference

The error happens every time.

**Expected behavior**
Popup should automatically be closed upon opening another menu

**Screenshots**
http://recordit.co/crmMr3V2Q3

**Desktop**

- OS: Windows 10
- BleachBit version 4.3.0

**Additional context**
Every popup on the menu section should be included in this ticket.

abitrolly:
Could you, please, add a clear and concise problem description to your reports? For example:

* I'm always frustrated when [...] because [...]
* I would like it if [...] happened when I [...] because [...]

banaagravrommel:
> Could you, please, add a clear and concise problem description to your reports? For example:
> 
> * I'm always frustrated when [...] because [...]
> * I would like it if [...] happened when I [...] because [...]

I changed the description hope you understand it more

rados:
We can make the pop up modal, like the Preferences.

az0:
> We can make the pop up modal, like the Preferences.

OK, the dialog can be modal. This will also fix #1184

az0:
I cherry-picked 0933d6785ff355a254c583914e50033a7f1e9bde from the Pythohn 3.10 branch into master, so this is fixed now. The dialog is now modal

PR #1450, opened 2023-01-22
#1183 Open the "Make Chaff" dialog modally so as to prevent opening of other dialogs while it is open.

sanjaykaria:
…


az0:
This looks good. Thank you.

Issue #1410, opened 2022-10-21
Fullscreen glitch with 2265 Python 3.10 build using natively installed vcruntime140.dll


kenrap:
After experiencing the bug in #1409, closing and opening Bleachbit again would automatically go in a fullscreen mode without showing titlebar and covering the Windows taskbar. I couldn't get out of it using keyboard shortcuts to unmaximize or tile the window. I was able to close Bleachbit by hitting the Windows (super) key to reveal the taskbar and right click on the task to finally close it.

**Steps To reproduce the behavior**
1. Do the steps mentioned in #1409.
2. Close and open Bleachbit again.

**Expected behavior**
Should open in the usual floating-mode window and not be in fullscreen mode.

**Desktop Specs**
OS: Windows 11 (via qemu-kvm using virt-manager to manage passthrough configurations)
BleachBit 4.4.2.2265

**Additional context**
Deleting the `%APPDATA%\BleachBit\bleachbit.ini` file restores Bleachbit back to the state with the first bug in #1409 and therefore repeats this cycle.

az0:
@kenrap 

Does this happen with the newer [build 2310](https://ci.bleachbit.org/?prefix=dl/4.4.2.2310-mkhon-python310/)?

sanjaykaria:
> I couldn't get out of it using keyboard shortcuts to unmaximize or tile the window. I

@kenrap  You can use Alt+F4 to close Bleachbit in full screen mode (on Windows)

sanjaykaria:
@az0 

I think this issue is related to the action of adding the 2 options "window_fullscreen" and  "window_maximized", to fix #204.  Somehow, the option window_fullscreen is getting set and the "fullscreen glitch" which is basically BB in fullscreen mode occurs.

There is an option(icon on right-top) to maximize BleachBit on the GUI, but there is no option to make it FullScreen( at least I am not aware of it ) on the GUI. Also, I think the fullscreen option is not needed, only the maximize window should suffice.

So, if we remove the "window_fullscreen" option and just use "window_maximized", it should suffice to store the window size and position.

az0:
@sanjaykaria 

Yes, I do not see a way to make the application go full screen, so remove the code related to it.

az0:
_Administrative note_

[...]

PR #1451, opened 2023-01-23
#1410 Removing the "window_fullscreen" option as the option "window_maximized" will suffice to store the window size and position.

sanjaykaria:


mkhon:
I believe "maximized" property was there so that the window remains maximized when the screen geometry has changed (resolution changed or external display is attached)

az0:
> I believe "fullscreen" property was there so that the window remains fullscreen when the screen geometry has changed (resolution changed or external display is attached)

I don't see any button, keyboard shortcut, or other way to initially set the fullscreen mode. There is also no way for the user to unset it, and I do not see what was triggering it for the user in #1410.

sanjaykaria:
> I believe "fullscreen" property was there so that the window remains fullscreen when the screen geometry has changed (resolution changed or external display is attached)

This is addressed by the "window_maximized" property. Fullscreen is also a maximized screen, but without the BB toolbar. 
I guess it is not much useful to work without the BB toolbar, besides there being no option/shortcut to make BB fullscreen.

Ref: https://stackoverflow.com/questions/40741208/what-is-the-difference-between-the-selenium-methods-maximize-and-fullscreen#:~:text=A%20maximized%20window%20is%20not,title%20bar%20is%20not%20displayed.

PR #1460, opened 2023-01-31
Update installer script and strings

bovirus:
@az0 

Update mnmain instalkelr script and language strings.

- Update main script
Change name oin top left main window (propgram + version)
Change order of exe file properteios variable (easier to see)
Add some variable instead fixed strings
Add description option for installation module

- Update list of strings
Add some strings (changes in main script)
Change order of the strings to group for use

Please check and merge

In which file is the string in redbox? I didn't find it to fix.

image

..


az0:
@bovirus 

While you work on this, please convert it to a draft. There is a button "convert to draft" on the top-right of the page. When you are done, change the status back and write a comment.

windows/bleachbit.nsi
```diff
@@ -1,493 +1,519 @@
...
+!ifdef packhdr
+  ;Using UPX path info from setup_py2exe.py ->
+  !packhdr "$%TEMP%\exehead.tmp" '"\upx\upx.exe" -9 -q "$%TEMP%\exehead.tmp"'
+!endif
+
+
+;--------------------------------
+;Include Modern UI
+
+  !include "MUI2.nsh"
```
az0, line 41: @bovirus 

It looks like the _whole file_ was deleted and replaced, so it makes it hard to see where the differences are. Normally we can see only the lines that were changed.

The way it looks here  can be caused by changing the end of line characters (CR/LF). Less often, it is caused by changing whitespace.

Please be careful with the editor, so the changes are easier to see. This will make it easier to review and faster to get merged.

bovirus: @az0

I search draft option but I didn't find it.
Is it possible to download the temporary installer created? How do that?

bovirus:
@az0 

Now the installer should OK.

If you can please merge and I will test new installer.

az0:
You can download the installer for pull request from https://ci.appveyor.com/project/az0/bleachbit/history

Choose a build, and then click on artifacts

bovirus:
@the installer is OK.

There is a problem that new installer not create program group and the line in Windows program list.

the last build tht is working is 4.4.2.2338.

bovirus:
@az0

How can make the test without uplaoding nsi every time?
I need an archive with all filse that try collect the .nsi script to produce the installer.

az0:
The portable zip has all the files needed to build the NSIS installer

PR #1465, opened 2023-02-14
Fixes #441

### LINKED ISSUE #141: HiDPI Screen
Thanks for the fantastic cleaner.

I recently switched to a high resolution screen and while most of the elements scale quite well the tick boxes are very small and a bit difficult to press. Would it perhaps be possible to scale them a bit?

Second very small suggestion, it would be great if you could disable the overlay scrollbar to fit with Ubuntu 16.04. Nothing crucial though.

Many thanks in advance.

### LINKED ISSUE #441: Make sure that DBs get compressed last of a rule set
Make sure that DBs get compressed last of a rule set.
Seems there is at least a problem with the German version, because "DB compress" is on top and therefor get done first (before deletion of the rest).

2019-02-12 19_04_58-bleachbit

sanjaykaria:
Sort the cleaners by Id (instead of localized display name) so that the sort sequence is consistent across all languages, and specifically the "Vacuum" option gets executed at the end.

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -440,9 +440,9 @@ class TreeInfoModel:
                                                               self.on_row_changed)
 
     def sort_func(self, model, iter1, iter2, _user_data):
-        """Sort the tree by the display name"""
-        value1 = model[iter1][0].lower()
-        value2 = model[iter2][0].lower()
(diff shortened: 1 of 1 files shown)


az0:
@sanjaykaria 

I will take a closer look. Thank you

PR #1467, opened 2023-02-15
Code Optimizations: CleanerML.py

haxerzin:
// default_vars //

- list comprehension performance improvement
- wowvars tuple converted to list for faster iterations

// list_cleanerml_files //

- tuple converted to list for faster iterations
- file list in dirs are generated using generator exper. for memory optimization (avoids creating large list of files in memory)
- os.path.splitext is used to avoid converting file extensions to lowercase
- os.path.join used to avoid looping over files to generate full path

### DIFF
--- a/bleachbit/CleanerML.py
+++ b/bleachbit/CleanerML.py
@@ -32,7 +32,11 @@ from bleachbit import Cleaner
 import logging
 import os
 import sys
+import stat
 import xml.dom.minidom
+
+from os import listdir
+from os.path import isfile, join, splitext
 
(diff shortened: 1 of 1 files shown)


az0:
> Seems to be multiple errors with other files and not specific to this pull request. Can't figure out the point of failure on appveyor. Can someone check please.

Other builds are not failing: it's just this PR. At a glance, the errors indicate that this code changed the logic.

Maybe split your PR into smaller pieces. This will help troubleshoot the error. Also, it makes it easier to review and merge.

PR #1488, opened 2023-05-15
Clean Thunderbird installed as flatpak

MauroGuida:
Hi, this pull request is for adding support for Thunderbird flatpak. I have done a little bit of testing and it seams to work fine on my side.

### DIFF
--- a/cleaners/thunderbird.xml
+++ b/cleaners/thunderbird.xml
@@ -33,6 +33,7 @@
     <value search="glob" os="linux">~/.thunderbird/*</value>
     <value search="glob" os="linux">~/.mozilla-thunderbird/*</value>
     <value search="glob" os="linux">~/.thunderbird/Profiles/*</value>
+    <value search="glob" os="linux">~/.var/app/org.mozilla.Thunderbird/.thunderbird/*</value> <!-- Flatpak -->
     <value search="glob" os="windows">%AppData%\Thunderbird\Profiles\*</value>
   </var>
   <option id="cache">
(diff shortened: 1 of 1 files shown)


cleaners/thunderbird.xml
```diff
@@ -33,13 +33,15 @@
...
     <value search="glob" os="linux">~/.thunderbird/Profiles/*</value>
+    <value search="glob" os="linux">~/.var/app/org.mozilla.Thunderbird/.thunderbird/*</value> <!-- Flatpak -->
     <value search="glob" os="windows">%AppData%\Thunderbird\Profiles\*</value>
   </var>
   <option id="cache">
     <label>Cache</label>
     <description>Delete the web cache, which reduces time to display revisited pages</description>
     <action command="delete" search="walk.all" path="$$profile$$/Cache/"/>
     <action command="delete" search="walk.all" path="~/.cache/thunderbird/"/>
+    <action command="delete" search="walk.all" path="~/.var/app/org.mozilla.Thunderbird/cache/thunderbird/"/> <!-- Flatpak -->
```
fbruand, line 44: Hello @MauroGuida , I wonder why we don't clean up this flatpak's cache folder directly? From my point of view, since flatpak creates a folder dedicated to the cache, we could clean it up directly in order to also eliminate other data present (which is the case with some flatpaks and, for example, this one has the "mesa_shader_cache" folder). In this case, we would aim for "~/.var/app/org.mozilla.Thunderbird/cache/" instead of "~/.var/app/org.mozilla.Thunderbird/cache/thunderbird/".

MauroGuida: Because I didn't think about it. You're right, when installed via the OS repository, those apps use the `.cache` folder, and BleachBit cleans cache files like `mesa_shader_cache`. So, we should do the same for Flatpak apps.

PR #1511, opened 2023-07-27
fix: typo miss

rewrite0w0:
fixed korean translate miss typo

### DIFF
--- a/po/ko.po
+++ b/po/ko.po
@@ -1610,7 +1610,7 @@ msgid ""
 "Run BleachBit with administrator privileges to improve the accuracy of "
 "overwriting the contents of files."
 msgstr ""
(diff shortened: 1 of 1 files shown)


az0:
Thank you for your contribution. 

In the future, could you please make translation changes using [Weblate](https://hosted.weblate.org/projects/bleachbit/main/)? It is designed for translations. Through the integration, it will create a commit here in GitHub without requiring the PR process.

I have not tested whether Weblate works bidirectionally. It seems like it should work, but I suggest checking whether the next Korean translation does not overwrite your commit.

az0:
P.S., I uploaded some new strings to translate. If you have some free time, [see Weblate](https://hosted.weblate.org/projects/bleachbit/main/ko/#overview).

PR #1513, opened 2023-08-06
Update org.bleachbit.BleachBit.desktop

baraeerriyachy:
Implemented icon resolution fix by adding StartupWMClass=bleachbit to the .desktop file. This ensures the correct icon is displayed in the application launcher instead of the default blue icon. Tested on my system (archlinux), and the icon now appears as expected.

### DIFF
--- a/org.bleachbit.BleachBit.desktop
+++ b/org.bleachbit.BleachBit.desktop
@@ -123,4 +123,5 @@ Icon=bleachbit
 Categories=System;FileTools;GTK;
 Keywords=cache;clean;free;performance;privacy;
 StartupNotify=true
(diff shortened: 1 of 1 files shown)


az0:
@baraeerriyachy 
I would like to compare this on different Linux systems. Do you have before and after screenshots? If the screenshot is in your clipboard, you can paste images here in GitHub using CTRL+V as if you were pasting text.

baraeerriyachy:
after
before

baraeerriyachy:
before
after

az0:
Thank you. I am writing up test instructions for the beta release, and I will point people to look at the images. "A picture is worth a thousand words"

PR #1535, opened 2023-09-03
Update deleting autofill data in Chrome-like browsers.

rados:


az0:
@rados  Thank you for working on this.

If a user has an old database version, probably because he uninstalled a Chrome-based browser before it updated the database, then the new code will keep erroring on the old database, so a solution would be to modify `__shred_sqlite_char_columns()` to check that ecah column exists like `__sqlite_table_exists()`.

Let me know the next time you can work on this, and depending on our schedules, I may work on it.

For now, I will ask some users (with the new database) to test this.

rados:
@az0 Yes, I've missed this case, thank you. According to my observations the whole tables 'local_addresses' and 'local_addresses_type_tokens' are part  of the new db, so it will be enough to check if the tables exist. I can push this change quickly.

PR #1547, opened 2023-09-21
Update NsisMultiUserLang.nsh

bovirus:
@az0 

Please check and merge,


az0:
The UNINSTALL_TEXT string came from your commit 87c432b04. What do you think about removing that sentence from other languages?

bovirus:
@az0 

We already tried to change "Uninstal" in a multilanguage way but with current script layout is not possible.

You cannot remove other languages because NSIS is not like Inno Setup (that when didn't fin a language sue english base),.

NSIS reqire a specfic strings also if is euqla to english.
You cabnot avoiud to define a string.

PR #1577, opened 2023-11-06
use a more compatible way to determine the session type

proletarius101:
Use the environment variable`XDG_SESSION_ID` to determine the session type instead of invoking a command. This variable is also populated by systemd-logind. However, it can be filled by the host system if bleachbit is running in a sandbox without systemd in the sandbox.

Fixes https://github.com/flathub/org.bleachbit.BleachBit/issues/13

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -728,10 +728,7 @@ def dnf_autoremove():
 
 def is_linux_display_protocol_wayland():
     assert(sys.platform.startswith('linux'))
-    result = General.run_external(['loginctl'])
-    session = result[1].split('\n')[1].strip().split(' ')[0]
-    result = General.run_external(['loginctl', 'show-session', session, '-p', 'Type'])
-    return 'wayland' in result[1].lower()
+    return os.environ['XDG_SESSION_TYPE'] == 'wayland'
 
 
 def root_is_not_allowed_to_X_session():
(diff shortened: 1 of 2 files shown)


fabiscafe:
This also misses an exception if the env var is not set. For example on `sudo bleachbit`. That actually works here in the current version. (I dont know if its supposed to work.)

```
  File "/usr/bin/bleachbit", line 40, in <module>
    bleachbit.Unix.is_display_protocol_wayland_and_root_not_allowed()
  File "/usr/share/bleachbit/Unix.py", line 743, in is_display_protocol_wayland_and_root_not_allowed
    bleachbit.Unix.is_linux_display_protocol_wayland() and
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/bleachbit/Unix.py", line 731, in is_linux_display_protocol_wayland
    return os.environ['XDG_SESSION_TYPE'] == 'wayland'
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^
  File "<frozen os>", line 679, in __getitem__
KeyError: 'XDG_SESSION_TYPE'
```

az0:
I commited a change to help: https://github.com/bleachbit/bleachbit/commit/0e710eb4186d967f048ed54bb6499b7469a26e0f

If you have an issue, open a new ticket

PR #1600, opened 2023-12-03
migrate where possible from external mock package to stdlib unittest.mock

eli-schwartz:
It was moved to the stdlib in python 3.3, and the PyPI version is a backport for python versions less than 3.3.

A couple of features from newer python versions that are greater than the minimum supported version are used. Add if/else guards to utilize the backport on older versions of python:
- assert_called_once was introduced in python 3.6
- Unstarted patchers were introduced in python 3.8


eli-schwartz:
Got CI to pass by conditionalizing the use of the stdlib based on python >=3.8.

az0:
@eli-schwartz 

Still working on this?

eli-schwartz:
Thank you for reminding me that this PR exists. :)

I'm happy to do whatever it takes to get it merged. Ten months ago it was ready for merge and the CI issue wasn't something I caused afaik, so I also couldn't fix it... I can rebase the PR and we can see if it passes these days.

eli-schwartz:
It passes. Yay?

az0:
I upgraded the Windows framework from Python 3.4 to 3.11, so we can drop support for older Python versions. Python 3.8 is EOL now.

@eli-schwartz  There are other `import mock` statements now. Do you want to take a look at those for a new PR?

eli-schwartz:
Hmm, commits like fc640243aefaf966199e54e8d2a4fdd757b863d4 confuse me greatly.

```diff
diff --git a/tests/TestOptions.py b/tests/TestOptions.py
index beea37c4..f82f9d90 100644
--- a/tests/TestOptions.py
+++ b/tests/TestOptions.py
@@ -22,12 +22,15 @@
 Test case for module Options
 """
 
+import errno
+import mock
+import os
+import unittest.mock
+
 from tests import common
 import bleachbit.Options
 from bleachbit import NoOptionError
 
-import os
-
 
 class OptionsTestCase(common.BleachbitTestCase):
```

There is certainly no need to import both of them. :)

eli-schwartz:
> I upgraded the Windows framework from Python 3.4 to 3.11, so we can drop support for older Python versions. Python 3.8 is EOL now.

I don't like to impose on other projects' support schedules. :) What versions do you want to support? e.g. python 3.9 is still supported until the end of the year, and Debian oldstable ships with python 3.9 and will be supported for ~10 months past the official EOL of python 3.9. Red Hat 9 is the most *recent* version, and will continue to ship with python 3.9 until they go out of maintenance support in 2032, but Red Hat 10 will be released "sometime mid-2025", so probably May or June.

Maybe most of this doesn't matter since people who install new versions of software only do so on the newest version of an operating system.

Anyways dropping <3.8 is easy enough.

az0:
unittest is needed only to run the tests, and when a user runs the application normally, it won't import unittest. Until Ubuntu 20.04 is EOL, I will continue to provide a .deb for it, but it doesn't ship with the tests/*py code. I currently provide a .deb for Ubuntu 18.04, but it is EOL, so I won't update it.

**Let's target Python 3.9 through 3.13.**

Issue #1658, opened 2024-04-09
Google Chrome bug


unicorn76:
Error: google_chrome.cache: Command to clean JSON file, path=C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\Preferences, address=dns_prefetching/host_referral_list 
Traceback (most recent call last):
  File "bleachbit\Worker.pyo", line 87, in execute
  File "bleachbit\Command.pyo", line 250, in execute
  File "bleachbit\FileUtilities.pyo", line 301, in clean_json
  File "json\__init__.pyo", line 268, in load
  File "json\__init__.pyo", line 314, in loads
ValueError: Unexpected UTF-8 BOM (decode using utf-8-sig)
Error: google_chrome.cache: Command to clean JSON file, path=C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\Preferences, address=dns_prefetching/startup_list 
Traceback (most recent call last):
  File "bleachbit\Worker.pyo", line 87, in execute
  File "bleachbit\Command.pyo", line 250, in execute
  File "bleachbit\FileUtilities.pyo", line 301, in clean_json
  File "json\__init__.pyo", line 268, in load
  File "json\__init__.pyo", line 314, in loads
ValueError: Unexpected UTF-8 BOM (decode using utf-8-sig)
Error: google_chrome.cache: Command to clean JSON file, path=C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\Preferences, address=net/http_server_properties/servers 
Traceback (most recent call last):
  File "bleachbit\Worker.pyo", line 87, in execute
  File "bleachbit\Command.pyo", line 250, in execute
  File "bleachbit\FileUtilities.pyo", line 301, in clean_json
  File "json\__init__.pyo", line 268, in load
  File "json\__init__.pyo", line 314, in loads
ValueError: Unexpected UTF-8 BOM (decode using utf-8-sig)
Clean file 0B C:\Users\unico\AppData\Local\Google\Chrome\User Data\Local State
Clean file 0B C:\Users\unico\AppData\Local\Google\Chrome\User Data\Local State
Clean file 0B C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\databases\Databases.db
Clean file 0B C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\Web Data
Clean file 0B C:\Users\unico\AppData\Local\Google\Chrome\User Data\Default\History
[...]

PR #1663, opened 2024-04-28
Open json file in encoding supporting BOM.

rados:


az0:
For cross reference, this was to fix https://github.com/bleachbit/bleachbit/issues/1658

Issue #1015, opened 2020-09-11
Turn off warnings...


anonymous1184:
Is there a hidden option or an undocumented command line switch to disable warnings? Or some way to enable something like "do not display this warning from now on..."?

az0:
Not at this time.

anonymous1184:
Hopefully it get implemented sometime in a near future as this is somewhat annoying. Thanks a lot.

az0:
@anonymous1184 
[BleachBit 5.0.2.3179-dev](https://ci.bleachbit.org/?prefix=dl/5.0.2.3179-dev/) adds a new feature that for each option-specific warning, the user can remember the choice to enable the option despite the warning. In the preferences, the user can reset the warnings. Try it out, and let me know how it works.

PR #1668, opened 2024-05-06
Fix 1626 - initial implementation, feedback needed.

rados:
I would like to show this initial implementation and collect feedback. The main question is about the warnings. For now I accumulate them per cleaner / section and show the accumulated ones, one by one. For example if there are two or more messages for given section (e.g. Firefox) only one is displayed, and so on for the rest of the sections. I have another suggestion. We can re-name the button to 'Force select all' and show a single message that describes what are the caveats in principle. Or we can have a setting in the preference dialog which toggles warnings on 'Select all'. Initial suggestion "An improvement would be to show all the warnings simultaneously. There are check boxes for each option with all options enabled by default." is very good but still I am not sure that additional set of checkboxes in the warning dialog would be effective enough.

### DIFF
--- a/bleachbit/GUI.py
+++ b/bleachbit/GUI.py
@@ -528,13 +528,13 @@ class TreeDisplayModel:
                                                _('Confirm'))
                 if Gtk.ResponseType.OK != resp:
                     # user cancelled, so don't toggle option
-                    return None, None
+                    return None
                 
             elif warning and all_toggled:
                 return msg, path
                 
         model[path][1] = value
-        return None, None
+        return None
         
 
     def col1_toggled_cb(self, cell, path, model, parent_window):
@@ -566,7 +566,7 @@ class TreeDisplayModel:
         msgs_paths = []
         while child:
             msg_path = self.set_cleaner(child, model, parent_window, is_toggled_on, all_toggled)
-            if msg_path is not (None, None):
+            if msg_path is not None:
                 msgs_paths.append(msg_path)
             child = model.iter_next(child)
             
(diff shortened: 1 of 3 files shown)


az0:
Thanks for working on this. For easy reference, I made a screen recording while using your code.

https://github.com/user-attachments/assets/be6e63a9-a4b3-4e5f-8da0-5e2f14f5ea6c

> For now I accumulate them per cleaner / section and show the accumulated ones, one by one. For example if there are two or more messages for given section (e.g. Firefox) only one is displayed, and so on for the rest of the sections

OK, I see that. I noticed that "select all" gave warnings for the deep scan section, even when I already had that enabled. Likewise, doing "select all" twice in a row repeats all the warnings, even though everything is enabled, but I'd like to **skip warnings for any enabled option**.

Another change: use new warning dialog when clicking a section (e.g., deep scan has warnings for each option). In other words, it's not just  the new "select all" button that uses the new dialog.

> Or we can have a setting in the preference dialog which toggles warnings on 'Select all'.

You mean a preference to always ignore warnings? That's like this issue https://github.com/bleachbit/bleachbit/issues/1015

>Initial suggestion "An improvement would be to show all the warnings simultaneously. There are check boxes for each option with all options enabled by default." is very good but still I am not sure that additional set of checkboxes in the warning dialog would be effective enough.

I am not sure why it would not be effective, and I am open to hearing more or seeing it. However, this is not urgent. I'm going to skip this change for the upcoming 5.0 release.


I made a standalone mockup
image
[mockup_warning_dialog_py.txt](https://github.com/user-attachments/files/19145049/mockup_warning_dialog_py.txt)

PR #1693, opened 2024-06-24
fix: missing developer name in the metainfo file

proletarius101:
The developer name field (https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer) is required in Flathub. The absence of it prevents updates.


az0:
@proletarius101 ok, tagging now

Issue #972, opened 2020-07-08
Consider change  `/usr/bin/env python3` to `/usr/bin/python3` in `/usr/bin/bleachbit` ?


Jerrywang959:
I use bleachbit 4.0.0 in deepin linux v20 beta. I type `bleachbit` in terminal, it shows that `ModuleNotFoundError: No module named 'gi'`.  Later, I find it's becaues I use python from anaconda. When I change the first line of `/usr/bin/bleach` from `#!/usr/bin/env python3` to `#!/usr/bin/python3`, it works. 

In my opinion, it's better to substitute `/usr/bin/python3` for  `/usr/bin/env python3` because a software needs just use python from system.

az0:
On my Ubuntu  20.04 system there are 129 files in `/usr/bin` with a shebang interpreter directive

 `grep -IPs "^\#" -m 1 *  |grep python   |wc -l`

Of those, only 3 use `env`, so they agree with you.

mkhon:
I don't think the fix was correct. See this for example: https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take

The user just had another python installation that he used as a default and he just needed to add gi packages there.

Jerrywang959:
> I don't think the fix was correct. See this for example: https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take
> 
> The user just had another python installation that he used as a default and he just needed to add gi packages there.

However, It hard for me to add gi packages for python from anaconda.
I had tried 

```shell
pip install gi
```

it shows

```python
ERROR: No matching distribution found for gi
```

The version of python in my envrionment is 3.7.6 and I use deepin Linux v20. So I can just only use `#!/usr/bin/python3` to start bleachbit

az0:
> I don't think the fix was correct.

There are a few things going on in the StackOverflow article. 
I am not using the unversioned `python` nor the `/usr/local/bin` directory.

So, you would undo change 4580240f7e86cd0732c89d5ab7dd77c4fc9b6864 , or do you mean something else?

mkhon:
bleachbit should use default Python (this is what "#!/usr/bin/env python3" does. Other users may have reasons to use non-system Python with bleachbit.

If the user has non-system Python that he can not use with Bleachbit for some reason he should run bleachbit as "PATH=/usr/bin:$PATH bleachbit".

PR #1714, opened 2024-09-10
Improve portability around hardcoded /usr/share in bleachbit.py; respect destdir and prefix in po/Makefile; correct shebangs

Zopolis4:
This will correct the hardcoded `/usr/share` path in the installed bleachbit.py to match the `/usr/local` or other prefix passed to the makefile. If the prefix is `/usr`, things will behave as before.

Also, have pass the values of prefix and datadir to po/Makefile and have it respect them.

Additionally, correct the shebangs to not be hardcoded to `/usr/bin`. I'm not sure what happened with #972, as the issues that were brought up there after 4580240f7e86cd0732c89d5ab7dd77c4fc9b6864 seem to have been ignored and the issue closed regardless. I have reverted that change, as well as fixed a few other broken shebangs-- I have not added shebangs to files that were missing them in the first place, only those that had them removed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -75,7 +75,7 @@ install:
 	$(INSTALL_DATA) bleachbit.png $(DESTDIR)$(datadir)/pixmaps/
 
 	# translations
-	make -C po install DESTDIR=$(DESTDIR)
+	make -C po install prefix=$(prefix) datadir=$(datadir) DESTDIR=$(DESTDIR)
 
 	# PolicyKit
 	mkdir -p $(DESTDIR)$(datadir)/polkit-1/actions
--- a/Makefile
+++ b/Makefile
@@ -43,6 +43,8 @@ install:
 	mkdir -p $(DESTDIR)$(bindir)
 	$(INSTALL_DATA) bleachbit.py $(DESTDIR)$(bindir)/bleachbit
 	chmod 0755 $(DESTDIR)$(bindir)/bleachbit
+	# Update the hardcoded /usr/share in bleachbit.py to match the datadir set by the user.
+	sed -i 's|/usr/share|$(datadir)|g' $(DESTDIR)$(bindir)/bleachbit
 
 	# application launcher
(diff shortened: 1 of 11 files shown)


bleachbit.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/python3
+#!/usr/bin/env python3
```
az0, line 1: @Zopolis4 
This commit has a mix of `#!/usr/bin/env python3` and #!/usr/bin/env python`. Let's use the former because only Python3 is supported.

Makefile
```diff
@@ -43,6 +43,8 @@ install:
 	mkdir -p $(DESTDIR)$(bindir)
 	$(INSTALL_DATA) bleachbit.py $(DESTDIR)$(bindir)/bleachbit
 	chmod 0755 $(DESTDIR)$(bindir)/bleachbit
+	# Update the hardcoded /usr/share in bleachbit.py to match the datadir set by the user.
+	sed -i 's|/usr/share|$(datadir)|g' $(DESTDIR)$(bindir)/bleachbit
```
az0, line 47: @Zopolis4 
There are many references to `/usr/share` in `__init__.py`, though not all should be changed. One solution is to make a Python variable with `/usr/share` that sed can modify.

`/usr/share/common-licenses/GPL-3` is an example that should not be modified.

Zopolis4: I'm only modifying `bleachbit.py` here, unless you mean there are further portability problems in `__init__.py` that I have not adressed?

az0: @Zopolis4 
Yes, search `__init__.py` for `/usr/share`.

az0:
@Zopolis4 
Is this change to help with packaging or installation for a certain Linux distribution?

Zopolis4:
Yes-- chromebrew carried a number of downstream patches to fix behavior around installing into /usr/local, and I am attempting to fix the issue upstream so we no longer need to carry those patches.

PR #1733, opened 2024-10-30
Copy To/Move To Dolphin history

mariovillani:
Clears recent locations where files have been copied or moved to (menu triggered when right clicking on a file in Dolphin).

Needs a fix since "ini" command parser expects to not have duplicates, so before cleaning up it is throwing: 

`configparser.DuplicateOptionError: While reading from '/home/myuser/.config/dolphinrc' [line 30]: option '1920x1080 screen' in section 'KPropertiesDialog' already exists`


az0:
@mariovillani 

Where is the menu? I right clicked ona file.

image

mariovillani:
Needs to be enabled in Dolphin Context Menù settings. "Copy in" and "Move to" commands, then will be visible right after "Activities" context menu entry.

PR #1774, opened 2025-02-16
fix: safely delete a file

zeze-zeze:
Open the parent directory of the file before removing it to avoid an attacker changing the parent directory to a junction.

### DIFF
--- a/bleachbit/Windows.py
+++ b/bleachbit/Windows.py
@@ -143,6 +143,23 @@ def csidl_to_environ(varname, csidl):
     # there is exception handling in set_environ()
     set_environ(varname, sppath)
 
+def delete_file(pathname, func_remove=None):
+    """Safely delete a file"""
+    handle = win32file.CreateFile(
+        os.path.dirname(os.path.abspath(pathname)),
+        win32file.GENERIC_READ,
+        win32file.FILE_SHARE_READ,
+        None,
+        win32file.OPEN_EXISTING,
+        win32con.FILE_FLAG_BACKUP_SEMANTICS,
+        None
+    )
+    if handle != win32file.INVALID_HANDLE_VALUE:
+        if func_remove:
+            func_remove(pathname)
+        else:
+            win32file.DeleteFile(pathname)
+        win32file.CloseHandle(handle)
 
 def delete_locked_file(pathname):
     """Delete a file that is currently in use"""
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -51,6 +51,8 @@ if 'nt' == os.name:
     os_path_islink = os.path.islink
     os.path.islink = lambda path: os_path_islink(
         path) or bleachbit.Windows.is_junction(path)
+    os_remove = os.remove
+    os.remove = lambda pathname: bleachbit.Windows.delete_file(pathname, func_remove=os_remove)
 
 if 'posix' == os.name:
     from bleachbit.General import WindowsError


az0:
Could you please check [this error on AppVeyor](https://ci.appveyor.com/project/az0/bleachbit/builds/51524961)?

```
Now you can run cd ..; python3 bleachbit/GUI.py
mingw32-make.exe: Leaving directory 'C:/projects/bleachbit/po'
%PYTHON_HOME%\python.exe -m windows.setup_py2exe
2025-02-16 11:47:42 - INFO - ROOT_DIR C:\projects\bleachbit
2025-02-16 11:47:42 - INFO - Getting BleachBit version
  File "C:\projects\bleachbit\x86-windows\tools\python3\lib\runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\projects\bleachbit\x86-windows\tools\python3\lib\runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "C:\projects\bleachbit\windows\setup_py2exe.py", line 666, in <module>
    import bleachbit
  File "C:\projects\bleachbit\bleachbit\__init__.py", line 202, in <module>
    from bleachbit import Windows
  File "C:\projects\bleachbit\bleachbit\Windows.py", line 41, in <module>
    from bleachbit import Command, FileUtilities, General
  File "C:\projects\bleachbit\bleachbit\Command.py", line 30, in <module>
    from bleachbit import FileUtilities
  File "C:\projects\bleachbit\bleachbit\FileUtilities.py", line 54, in <module>
    os.remove = bleachbit.Windows.delete_file
AttributeError: partially initialized module 'bleachbit' has no attribute 'Windows' (most likely due to a circular import)
Command exited with code 1
```

zeze-zeze:
I've force pushed another commit to this PR, but it seems to keep processing updates. Not sure if it's stuck for some reasons.

az0:
> I've force pushed another commit to this PR, but it seems to keep processing updates. Not sure if it's stuck for some reasons.
> 
> [zeze-zeze@cd731a2](https://github.com/zeze-zeze/bleachbit/commit/cd731a2caddf3d6bb0e244e43f617b89fb2bc162)

Yes, I see "Recent push is still being processed, and will show up in a bit"

bleachbit/Windows.py
```diff
@@ -144,21 +143,31 @@ def csidl_to_environ(varname, csidl):
...
+        win32file.GENERIC_READ,
+        win32file.FILE_SHARE_READ,
+        None,
+        win32file.OPEN_EXISTING,
+        win32con.FILE_FLAG_BACKUP_SEMANTICS,
+        None
+    )
+    if handle != win32file.INVALID_HANDLE_VALUE:
+        if func_remove:
+            func_remove(pathname)
```
az0: if func_remove() throws an exception, then this needs CloseHandle to avoid a leak.

bleachbit/Windows.py
```diff
@@ -144,21 +143,31 @@ def csidl_to_environ(varname, csidl):
...
+    handle = win32file.CreateFile(
+        os.path.dirname(os.path.abspath(pathname)),
+        win32file.GENERIC_READ,
+        win32file.FILE_SHARE_READ,
+        None,
+        win32file.OPEN_EXISTING,
+        win32con.FILE_FLAG_BACKUP_SEMANTICS,
+        None
+    )
+    if handle != win32file.INVALID_HANDLE_VALUE:
```
az0: If current user doesn't have admin rights, can we skip the new methods (like calling CreateFile)? This could reduce risk of bugs in edge cases in new code paths. Also, it avoids any potential performance penalty for CreateFile(), especially avoidable cases with many files in a single directory.

Could share flags WRITE and DELETE be granted (with READ) to avoid denying valid access to other processes?

In case CreateFile() fails, this silently skips: no message or further attempt at deletion. Minimally, we need to notify the user.

If the current user doesn't have ability to succeed with CreateFile(), would an attack have the ability to change the directory to a junction? If so, then we could safely fall back to normal delete.

az0 (commented):
Thanks for reporting the problem, providing a solution, and for your patience.  

Please see my review.

It's been a long time since you submitted this, so let me know whether you want to continue with this PR or whether you want to pass it off to someone else.

az0:
I took a similar but alternate approach, and binary is available here for testing
https://ci.bleachbit.org/?prefix=dl/6.0.0.3518-dev/

Thanks!

bleachbit/Windows.py
zeze-zeze: > If current user doesn't have admin rights, can we skip the new methods (like calling CreateFile)?

Attacker can still use the trick to delete other users' files which the attacker doesn't have privilege to access.

> Could share flags WRITE and DELETE be granted (with READ) to avoid denying valid access to other processes?

No. Here we use CreateFile instead of os.remove is because we need to restrict the WRITE and DELETE access from other process.

> If the current user doesn't have ability to succeed with CreateFile(), would an attack have the ability to change the directory to a junction? If so, then we could safely fall back to normal delete.

Yes. The reason why we need CreateFile is because it restricts the access right from other process. If we let other process WRITE or DELETE the file, an attacker can change the file to delete to a symbolic link.

PR #1874, opened 2025-07-02
Add description for Exaile and update Audacious description

TheTimeTombs:
Add missing description for Exaile.

### DIFF
--- a/cleaners/audacious.xml
+++ b/cleaners/audacious.xml
@@ -21,7 +21,7 @@
 -->
 <cleaner id="audacious" os="linux">
   <label>Audacious</label>
-  <description>Audio player</description>
+  <description>Media player</description>
   <option id="cache">
     <label>Cache</label>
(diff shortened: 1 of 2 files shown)


az0:
Good catch that it is missing.

What do you think about "Media player" instead of "Music player"? The former is already translated to many languages. Here is the [translation template](https://github.com/bleachbit/bleachbit/blob/master/po/bleachbit.pot).

If you do that, please also change Audacious to "Media player," so we do not need to translate an extra string.

TheTimeTombs:
> Good catch that it is missing.
> 
> What do you think about "Media player" instead of "Music player"? The former is already translated to many languages. Here is the [translation template](https://github.com/bleachbit/bleachbit/blob/master/po/bleachbit.pot).
> 
> If you do that, please also change Audacious to "Media player," so we do not need to translate an extra string.

That sounds like a good idea. I will implement that change.

PR #1879, opened 2025-07-09
Create vivaldi.xml

ghost:
Vivaldi cache cleaniring configuration added (I'm sorry for the xml I published before. I made this mistake without targeting the platform. Anyway, this code will work on all platforms.)

### DIFF
--- a/vivaldi.xml
+++ b/vivaldi.xml
@@ -1,0 +1,71 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<cleaner id="vivaldi">
+  <label>Vivaldi</label>
+  <description>Deep clean for Vivaldi browser (Windows, Linux, macOS)</description>
+
+  <!-- Detect running process -->
+  <running type="exe" os="windows" same_user="true">vivaldi.exe</running>
+  <running type="exe" os="linux" same_user="true">vivaldi</running>
+  <running type="exe" os="macos" same_user="true">Vivaldi</running>
+
+  <!-- Base configuration directory -->
+  <var name="base">
+    <value os="windows">%LOCALAPPDATA%\Vivaldi\User Data</value>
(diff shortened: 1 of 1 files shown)


vivaldi.xml
```diff
@@ -0,0 +1,71 @@
...
+    <action command="delete" search="walk.files" path="$$profile$$/GPUCache"/>
+    <action command="delete" search="walk.files" path="$$profile$$/ShaderCache"/>
+    <action command="delete" search="walk.files" path="$$profile$$/GrShaderCache"/>
+    <action command="delete" search="walk.files" path="$$profile$$/GraphiteDawnCache"/>
+  </option>
+
+  <!-- Cookies -->
+  <option id="cookies">
+    <label>Cookies</label>
+    <description>Delete cookies and associated files</description>
```
az0: Please check all the strings against other browsers to avoid unneeded re-translation.

az0 (changes requested):
Thanks for your patience. Please see my review. 

Here is a more general note: Many Chromium-based browsers like Vivaldi can be setup easily by starting with chromium.xml or google_chrome.xml and modifying just a few lines:

* cleaner ID attribute
* cleaner LABEL
* VAR 
* RUNNING

This approach would avoid the translation issue noted, plus clean more, while also using standard option ID values.

I know it's been a while since you submitted in July. Let me know if you want to want to continue with this PR or want someone else to take the next steps.

ghost:
Hello, I've received the necessary changes, but when I first started this project, the Bleachbit XML structure was quite different, that's why I wrote it this way. And yes, I want it to continue. I'm the maintainer.

ghost:
@az0 Hello, I've made the necessary changes, you can check them out. This new version is now fully supported and compatible.

PR #2005, opened 2026-02-09
Update chromium.xml

Mr-Update:
- Entries added based on (Ungoogled) Chromium 144 in Windows 10/11

### DIFF
--- a/cleaners/chromium.xml
+++ b/cleaners/chromium.xml
@@ -49,7 +49,6 @@
     <action command="delete" search="file" path="$$profile$$/Network Persistent State"/>
     <action command="delete" search="file" path="$$profile$$/Network/Network Persistent State"/>
     <action command="delete" search="walk.all" path="$$base$$/ShaderCache"/>
-    <action command="delete" search="walk.all" path="$$base$$/ShaderCache/GPUCache"/>
     <action command="delete" search="walk.all" path="$$base$$/GrShaderCache"/>
     <action command="delete" search="walk.all" path="$$base$$/GraphiteDawnCache"/>
     <action command="delete" search="walk.all" path="$$profile$$/File System"/>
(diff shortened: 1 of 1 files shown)


cleaners/chromium.xml
```diff
@@ -64,6 +69,8 @@
     <action command="delete" search="glob" path="$$base$$\B*.tmp"/>
     <action command="delete" search="walk.all" path="$$profile$$\Default\Application Cache\"/>
     <action command="delete" search="walk.files" path="$$profile$$\Cache\"/>
+    <action command="delete" search="walk.files" path="$$profile$$\Cache\Cache_Data\"/>
```
az0: `$$profile$$\Cache\` above includes `$$profile$$\Cache\Cache_Data\`  because `walk.all` includes the subdirectories. As it, it will preview twice, but can delete only once. Please remove the extra line.

cleaners/chromium.xml
```diff
@@ -64,6 +69,8 @@
...
+    <action command="delete" search="walk.files" path="$$profile$$\Code Cache\js\"/>
```
az0: How about removing `js` so it includes all of Code Cache?

cleaners/chromium.xml
```diff
@@ -47,11 +47,16 @@
     <description>Delete the web cache, which reduces time to display revisited pages</description>
     <action command="delete" search="file" path="$$base$$/Safe Browsing Channel IDs-journal"/>
     <action command="delete" search="file" path="$$profile$$/Network Persistent State"/>
+    <action command="delete" search="file" path="$$profile$$/Network/Network Persistent State"/>
     <action command="delete" search="walk.all" path="$$base$$/ShaderCache"/>
+    <action command="delete" search="walk.all" path="$$base$$/ShaderCache/GPUCache"/>
```
az0: `$$base$$/ShaderCache/GPUCache` is not needed because the line above includes it.

az0 (changes requested):
Thank you for these changes. Please see the review comments.

az0:
@Mr-Update 
Thank you.

While checking these changes on Ungoogled Chromium 144 on Windows, some files you listed were not created initially , so I improved the [Finding Files to Delete documentation](https://docs.bleachbit.org/cml/finding-files-to-delete.html) with tips to generate the files.

Issue #2034, opened 2026-02-25
Blank GUI window and Rendering Deadlock due to interrupted cleaner loading


banaagravrommel:
Describe the bug
The application launches to a blank/white screen where the window frame is visible, but no internal elements (sidebar, cleaner list, or buttons) are rendered. This appears to be a rendering deadlock caused by a failed cleaner loading sequence that is caught by a Try/Except block but fails to return a valid fallback value to the UI thread.

To Reproduce
1. Open the terminal on Debian 13.
2. Create a malformed cleaner file with invalid UTF-8 sequences by running:
mkdir -p ~/.config/bleachbit/cleaners
echo -e "<?xml version=\"1.0\" encoding=\"UTF-8\"?>\n<cleaner id=\"poison\">\n<label>\xFF\xFE\xFD Broken Encoding</label>\n</cleaner>" > ~/.config/bleachbit/cleaners/broken.xml
3. Launch BleachBit.
4. Observe the main window appearing as a blank, non-responsive frame.

The error happens every time the malformed XML is present.

Expected behavior
The application should skip the broken.xml file, log an encoding error to the console, and proceed to render the GUI for all other valid cleaners. A single malformed third-party cleaner should not prevent the entire application interface from loading.

Screenshots

https://github.com/user-attachments/assets/0af34ce1-53b0-4889-90fd-ecc0a39bf15e

Desktop (please complete the following information):
OS: Debian 13 (Trixie)
BleachBit version: 5.1.0

Additional context
This is a regression in the fix for UnicodeDecodeError. While the 5.1.0 update prevents a hard crash via a Try/Except block, it creates a "silent failure" state. Because the cleaner loading thread encounters an error and likely returns an incomplete data structure (or None), the GTK rendering engine hangs while waiting for the cleaner list to populate the sidebar.

There is a high probability this is linked to the NameError: name 'bleachbit' is not defined bug; if the exception handler tries to use the bleachbit module to log the encoding error, it triggers a secondary crash that deadlocks the UI thread.

PR #2066, opened 2026-03-20
Fix GUI deadlock when loading malformed XML cleaners (#2034)

chidoziemanagwu:
# Pull Request: Fix GUI deadlock when loading malformed XML cleaners (#2034)

**Context:**
Currently, if a user has a malformed XML file (e.g., containing invalid UTF-8 byte sequences) in their `~/.config/bleachbit/cleaners/` directory, BleachBit fails to start properly. The main GUI window opens but remains completely blank and unresponsive. This acts as a denial-of-service vulnerability against the application: a single anomalous cleaner file entirely prevents the user from accessing their privacy tools. This PR resolves [Issue #2034](https://github.com/bleachbit/bleachbit/issues/2034).

**The Approach:**
The root cause existed at two interconnected levels within `CleanerML.py`. I have implemented targeted exception handling and generator recovery to fix both.

**1. Broadening Exception Handling in `CleanerML.__init__()`**
Previously, the constructor only caught XML-specific errors (`ParseError` and `ExpatError`). I broadened this block to catch a generic `Exception`. This ensures that native OS and encoding errors from `ElementTree.parse()` (such as `UnicodeDecodeError` or `FileNotFoundError`) are gracefully caught, logged, and result in a safe, unusable cleaner object rather than corrupting the call stack.

```python
# bleachbit/CleanerML.py
try:
    tree = xml.etree.ElementTree.parse(pathname)
    root_element = tree.getroot()
except Exception as e: # <--- Changed from: except (xml.etree.ElementTree.ParseError, xml.parsers.expat.ExpatError)
    logger.error("Error parsing CleanerML file %s with error %s", pathname, e)
    return
```

**2. Generator Recovery in `CleanerML.load_cleaners()`**
When an exception was caught during cleaner instantiation, the loop `continue`d but skipped the necessary GTK progress updates. I added the increment and `yield True` steps inside the error handler so that GTK's idle loop (`GLib.idle_add`) remains responsive and the `register_cleaners` generator successfully yields control back to the UI, preventing the deadlock.

```python
# bleachbit/CleanerML.py
try:
    xmlcleaner = CleanerML(pathname)
except Exception:
    logger.exception(_("Error reading cleaner: %s"), pathname)
    files_done += 1                               # <--- Added
    cb_progress(1.0 * files_done / total_files)   # <--- Added
    yield True                                    # <--- Added
    continue
```

**3. Test Coverage (`TestCleanerML.py`)**
I added two new unit tests to specifically simulate the `\xff\xfe\xfd` invalid UTF-8 byte sequences reported in the issue, ensuring both the constructor and the loader pipeline handle them gracefully without crashing.

```python
def test_load_cleaners_invalid_utf8(self):
    """Unit test for load_cleaners() with invalid UTF-8 encoding"""
    pcd = bleachbit.personal_cleaners_dir
    bleachbit.personal_cleaners_dir = self.mkdtemp(prefix='bleachbit-cleanerml-utf8')
    self.write_file(os.path.join(bleachbit.personal_cleaners_dir, 'broken_encoding.xml'),
                    contents=b'<cleaner id="poison">\n\xff\xfe\xfd Broken\n')
    list(load_cleaners())
    shutil.rmtree(bleachbit.personal_cleaners_dir)
    bleachbit.personal_cleaners_dir = pcd
```

**Impact:**
This makes BleachBit significantly more resilient. Corrupted, maliciously crafted, or strictly invalid personal cleaner files will now simply throw a logged error in the console, while the rest of the application loads and functions normally. All unit tests (`python -m pytest tests/TestCleanerML.py`) pass successfully.

### DIFF
--- a/bleachbit/CleanerML.py
+++ b/bleachbit/CleanerML.py
@@ -152,7 +152,7 @@ class CleanerML:
         try:
             tree = xml.etree.ElementTree.parse(pathname)
             root_element = tree.getroot()
-        except (xml.etree.ElementTree.ParseError, xml.parsers.expat.ExpatError) as e:
+        except Exception as e:
             logger.error(
                 "Error parsing CleanerML file %s with error %s", pathname, e)
             return
@@ -394,6 +394,9 @@ def load_cleaners(cb_progress=lambda x: None):
             # TRANSLATORS: Error message printed to the log.
             # %s expands to the path of the XML cleaner file
             logger.exception(_("Error reading cleaner: %s"), pathname)
+            files_done += 1
+            cb_progress(1.0 * files_done / total_files)
+            yield True
             continue
         cleaner = xmlcleaner.get_cleaner()
         if cleaner.is_usable():
--- a/tests/TestCleanerML.py
+++ b/tests/TestCleanerML.py
@@ -108,6 +108,26 @@ class CleanerMLTestCase(common.BleachbitTestCase):
         shutil.rmtree(bleachbit.personal_cleaners_dir)
         bleachbit.personal_cleaners_dir = pcd
 
+    def test_load_cleaners_invalid_utf8(self):
+        """Unit test for load_cleaners() with invalid UTF-8 encoding"""
+        pcd = bleachbit.personal_cleaners_dir
+        bleachbit.personal_cleaners_dir = self.mkdtemp(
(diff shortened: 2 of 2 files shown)


az0:
Thank you, I will take a closer look.

PR #2133, opened 2026-05-26
add support of multiple profile on chrome and edge cleaners

### LINKED ISSUE #1786: Google Chrome cleaner don't clean multiple profiles but only the Default one
Seems that selecting google_chrome.cache clear cache files only of "Default" profile (C:\Users\Fabio\AppData\Local\Google\Chrome\User Data\Default) but not other profiles (if present), for example I another profile (C:\Users\Fabio\AppData\Local\Google\Chrome\User Data\Profile 2) and bleachbit don't have deleted its cache.
I saw many users with multiple profiles, on one today I saw also 4 profiles, it would be good to be able to have the cleaning function also on those who have multiple profiles.

 - OS: Windows 10
 - BleachBit version 4.6.2

Fantu:
Fixes #1786

### DIFF
--- a/cleaners/google_chrome.xml
+++ b/cleaners/google_chrome.xml
@@ -21,11 +21,17 @@ later.  See the COPYING file in the top-level directory.
   </var>
   <var name="profile">
     <value os="windows">%LocalAppData%\Google\Chrome\User Data\Default</value>
+    <value search="glob" os="windows">%LocalAppData%\Google\Chrome\User Data\Profile *</value>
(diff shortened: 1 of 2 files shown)


Fantu:
tested on windows 2019 and mint 22 with chrome with multiple profiles

Fantu:
Thanks for merge it, I did it after saw on production server (a windows terminal server) that with normal user (not administrator) bleachbit_console give error and in one case also crash on winapp2 browser cache, probably related to chrome temp system folder where normal user don't have permission.
On that servers anyway I still have version 5, I need to check also version 6

PR #2163, opened 2026-07-01
Remove non-POSIX es_419

PPN-SD:
The locale es_419 was added as an ISO language code https://github.com/bleachbit/bleachbit/commit/618fc7e96b25d97a2411989e0372a0410abb7a6c, but it is not part of ISO‑3166 (419 != country) and therefore not a valid POSIX locale. If es_149 appears in /etc/locale.alias or /usr/share/locale, it will be rejected by assertIsLanguageCode because it's not a POSIX-locale.

Since this locale was not added for the purpose of testing a non‑POSIX value, remove it for now.

See downstream bug: https://bugs.gentoo.org/978295

TODO: test only locales in native_locale_names, not everything installed in /usr/share/locale

### DIFF
--- a/bleachbit/Language.py
+++ b/bleachbit/Language.py
@@ -84,7 +84,6 @@ native_locale_names = \
      'en_US': 'United States English',
      'eo': 'Esperanto',
      'es': 'Español',
-     'es_419': 'Latin American Spanish',
      'et': 'eesti',
      'eu': 'euskara',
      'fa': 'فارسی',


az0:
That's interesting timing. The commit  618fc7 you referenced is from APril 2019, while two days ago I made commit 728f6dc3 to add es_419 to assertIsLanguageCode().

I agree it's not an official locale, but recognizing es_419 enables deleting `/usr/share/locale/es_419/LC_MESSAGES/pacman.mo` as seen on Arch Linux.

I agree that `native_locale_names` should not have es_419.

I didn't know about Gentoo Tinderbox: that's an interesting project. BleachBit has a mix of "real world" and mocked tests, and I could add a flag to disable the tests like this one that are sensitive to the environment. However, in this case, the test failure led to a useful change.

Thanks!

PPN-SD:
> That's interesting timing. 

Indeed!

> I could add a flag to disable the tests like this one that are sensitive to the environment. However, in this case, the test failure led to a useful change.

Yes, this test will be ignored with a basic sed command, like [some others](https://github.com/gentoo/gentoo/blob/fca2c6711a15e82d374541147a9162193d5a88d0/sys-apps/bleachbit/bleachbit-6.0.0.ebuild#L54) that cannot be passed in a sandboxed-build-env.

PR #2168, opened 2026-07-08
Persist zoom level across restarts

### LINKED ISSUE #2167: UI zooming not being persisted between app reruns
**Describe the bug**
As the title says

**To Reproduce**
1. Open BleachBit
2. Press <kbd>Ctrl</kbd> mouse scroll a few times to zoom in
3. Close BleachBit and reopen BleachBit
4. The UI zoom level is lost

The error happens every time.

**Expected behavior**
I would expect the UI level to be saved and persisted across application reruns

**Screenshots**

 - OS: Windows 10, Windows 11, Fedora 44
 - v6.0.2

**Additional context**
~~Add any other context about the problem here.~~

XhmikosR:
I only tested this on my Fedora machine and it seems to work.

Fixes #2167

### DIFF
--- a/bleachbit/GuiWindow.py
+++ b/bleachbit/GuiWindow.py
@@ -118,6 +118,8 @@ class GUI(Gtk.ApplicationWindow):
         self.textview.connect("scroll-event", self.on_scroll_event)
         self.connect("key-press-event", self.on_key_press_event)
         self._font_css_provider = None
+        if options.has_option("window_font_size"):
+            self.set_font_size(absolute_size=options.get("window_font_size"))
 
     def populate_window(self):
         """Create the main application window"""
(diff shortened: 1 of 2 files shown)


XhmikosR:
Arguably, if we had proper fractional scaling on Windows, this wouldn't matter a lot.

az0:
Good, thanks!

**CTRL+0** could clear the default instead of storing it.

By the way, I noticed a pre-existing issue that if TypeError is raised, then `gtk_font_name` has no value.

XhmikosR:
Yeah, I have that fixed in another branch, but I need the CI to run for non-first time contributor's forks. I have fixed most of the static analysis issues, but they need review from another pair of eyes :)

I will open PRs as we go.

PR #2170, opened 2026-07-08
Add CodeQL workflow

XhmikosR:
Supersedes #1432.

@az0: i went with the default queries. The one I personally use (`+security-and-quality`) results in too many issues for this repo, many of them are in tests. I left it commented it out for now.

I will try to address the issues after this is merged. :)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,59 @@
+name: CodeQL
+
+on:
+  push:
+    branches:
+      - master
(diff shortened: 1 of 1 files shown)


az0:
Thank you~

For follow up:

The `dependabot` exclusion can be removed  `master` is explicitly listed, and dependabot targets master.

It would be helpful to add `paths-ignore` like `test_ubuntu.yaml` to reduce runner resources.

infamous28 (commented):
D

PR #2171, opened 2026-07-08
Update and tighten GitHub workflows

XhmikosR:
* update actions to the latest versions
* add permissions
* add workflow_dispatch
* add `persist-credentials: false`
* cache pip
* add fail-fast: false
* guard translation updates and coverage to the main repo only
* format
* add dependabot config for action updates (monthly)

Non-whitespace diff: https://github.com/bleachbit/bleachbit/pull/2171/changes?w=1

Related #2170

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  - package-ecosystem: 'github-actions'
+    directory: '/'
+    cooldown:
+      default-days: 7
+    schedule:
+      interval: 'monthly'
+    groups:
+      github-actions:
+        patterns:
+          - '*'
--- a/.github/workflows/test_ubuntu.yaml
+++ b/.github/workflows/test_ubuntu.yaml
@@ -11,7 +11,7 @@ on:
       - 'po/**'
       - 'README.md'
       - 'windows/**'
-  pull_request:  # Run on all pull requests
+  pull_request: # Run on all pull requests
     paths-ignore:
       - 'bleachbit.spec'
       - 'debian/**'
@@ -19,16 +19,21 @@ on:
       - 'po/**'
       - 'README.md'
       - 'windows/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
 
 # 2025-05-01: ubuntu-latest is currently 24.04, and its native
 # python version is 3.12.
 # native python for 22.04 is 3.10.
 jobs:
   test:
+    name: test (${{ matrix.python-version }})
     timeout-minutes: 25
     strategy:
+      fail-fast: false
       matrix:
-        python-version: ['3.9', '3.13', '3.14']
         include:
           - python-version: '3.9'
             ubuntu-version: ubuntu-22.04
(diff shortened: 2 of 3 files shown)


az0:
I see push 22 minutes ago, so I assume this is a work in progress and will convert it to a draft.

XhmikosR:
Nah, I just added `workflow_dispatch` so that people with the proper access can trigger the workflow manually, if needed. It's ready.

XhmikosR:
I add `fail-fast: false` so that the other matrix jobs run too if a failure is detected. As for the failures, do you have any ideas why this happens? Is it new or maybe does it happen on PRs outside of the repo only?

Also, what are your thoughts about switching Windows CI to GitHub actions instead of AppVeyor?

Just trying to make CI more robust.

az0:
> I add `fail-fast: false` so that the other matrix jobs run too if a failure is detected. 

OK.

>As for the failures, do you have any ideas why this happens? Is it new or maybe does it happen on PRs outside of the repo only?

Do you mean this failure?

https://github.com/bleachbit/bleachbit/actions/runs/28994486720/job/86045252872

_Waiting for job to be queued...
Waiting for job to be queued...
Waiting for job to be queued...
Timed out waiting for job to be queued or build not found.
Failure._


> Also, what are your thoughts about switching Windows CI to GitHub actions instead of AppVeyor?

It would not be a small change, but I would review a PR. If you try it, please keep the PR focused to make it easier to review and merge. (If you have other changes in mind, please separate the PRs.)

In AppVeyor, I am considering changing the deployment from S3 to Cloudflare.

> Just trying to make CI more robust.

XhmikosR:
Yeah that failure. I noticed that Python 3.9 workflow did fail on my fork too randomly too.

All right, let's try to land the open PRs first and then I will try to work on the other things. I just need this and/or the #2170 merged so that you don't need to approve all workflow runs.

az0:
I can't follow every commit here because so many, but it looks  this change became much more complex with the appveyor.yml removal/replacement. I'd appreciate if you could keep the PRs more small and focused to help them get reviewed and merged please.

XhmikosR:
@az0: Sorry, I accidentally pushed the AppVeyor changes to this branch. I reverted to what it was.

XhmikosR:
@az0: here's my AppVeyor to GitHub Actions branch runs https://github.com/XhmikosR/bleachbit/actions?query=branch%3Axmr%2Frm-appveyor-gha

I don't think I can make the branch a lot smaller. The extents test might be moot on GHA. I will open this tomorrow, let me know if you have any suggestions in the meantime :)

az0:
> here's my AppVeyor to GitHub Actions branch runs

Thanks for working on that!

> I don't think I can make the branch a lot smaller. 

Yes, it can't be a small change. There are follow-up opportunities. With AppVeyor I recently tested that `pytest -n auto` did not improve performance, but it might be faster on GHA with 4 CPUs.  


This is my initial feedback

## BUILD_MODE

BUILD_MODE looks like it is always "fast" because an empty string is false.

`BUILD_MODE: ${{ (github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/'))) && '' || 'fast' }}`

This is not tested, but consider this

`BUILD_MODE: ${{ !(github.event_name == 'push' && (github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/'))) && 'fast' || '' }}`

## coveralls-finish

`coveralls-finish` is a nice replacement for the old method of polling across systems. However, I am not sure the `if` condition will behavior correctly in case test or windows fails or is skipped

This is not tested, but this might avoid hang waiting for test results

if: ${{ !cancelled() && github.repository == 'bleachbit/bleachbit' }}  

## paths-ignore

`paths-ignore` is missing some changes I made today in 7ffab012b2fea1333d55db432094418240a74731

az0:
FYI (no changes needed for `tests.yml`): I have a branch for a wxWidgets GUI that later will be merged into master, and I thought through how it would fit into `tests.yml`.  Because Windows+GTK requires special 32-bit Python 3.12 but wxWidgets runs on vanilla 64-bit Python, I will rename your `windows` job to `windows-gtk` and add a `windows-wx` job to run in parallel.

XhmikosR:
I rebased the branch and I'll work on cleaning it up a bit more later or in the next days. I'll make a PR after that :)

BTW I think that after the switch we should look into stop using MSYS stuff and eventually on 64-bit binaries for Windows. It will duplicate the CI/build time though, so maybe we should just drop 32-bit. Haven't had a look at this yet, just thinking out loud.

az0:
>maybe we should just drop 32-bit. 

I agree that long term I would like to drop 32-bit. Over the last few years, it has been a large effort to get GTK working on Windows with Python. For example, minor updates like within Python 3.12.x (which I am working on now) have been painful. It's enough work that it's [literally its own project](https://github.com/bleachbit/pygtkwin), and that's just for 32-bit, so I haven't started 64-bit GTK 3 or any kind of GTK 4.

The TUI and wxWidgets front ends, which are 64-bit on Windows, avoid these headaches.

---

Another FYI (no need to complicate your pending PR).

I am considering a change

- current: the Python version matrix of three versions, all using ubuntu-22.04 image
- future: ubuntu-24.04 and ubuntu-26.04 images with their native Python versions

Considerations:

- Using native Python versions better tests what users normally do.
- [ubuntu-22.04](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md) "will begin deprecation on September 17th and will be fully unsupported by April 17th"
- Python 3.9 is EOL as of October 2025
- CentOS 9 includes Python 3.9 by default,but CentOS 9 EOL is May 2027

Note to self: [your tests.yml file](https://raw.githubusercontent.com/XhmikosR/bleachbit/refs/heads/xmr/rm-appveyor-gha/.github/workflows/tests.yaml)

---

Plus, there will be another change to GHA for macOS. It has some failing modules, but there is progress.

XhmikosR:
One step at a time :)

Let's first finish with the GHA migration and then we look at the other changes. Personally, I'd keep the environments as close as possible to the upstream actions to save time and complexity when possible.

PR #2175, opened 2026-07-09
Avoid tempfile.mktemp in the symlink delete test

XhmikosR:
This along with #2174 and #2171 should fix the current CodeQL issues.

Later, we can decide if it's worth switching to a tighter CodeQL queries suite.

PR #2189, opened 2026-07-11
Tighten CodeQL and fix issues

XhmikosR:
The two Actions' issues will be fixed by #2171. The insecure temporary file in tests should be fixed by #2175.

The new issue about "Overly permissive file permissions" in tests should be ignored in the Security tab -> Code scanning -> click the alert -> Dismiss -> reason "Used in tests".

@az0: if you want me to change the branches CodeQL runs, let me know.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -3,7 +3,8 @@ name: CodeQL
 on:
   push:
     branches:
-      - master
+      - '**'
+      - '!dependabot/**'
     paths-ignore:
       - 'appveyor.yml'
       - 'bleachbit.png'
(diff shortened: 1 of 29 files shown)


XhmikosR:
@az0: this is what I suggest: https://github.com/XhmikosR/bleachbit/commit/8b4401cb5d1d59965d3d247e6e95500e8b9b8f1c

Let me know if you agree so that I pick those here.

XhmikosR:
I pushed my CodeQL branch with the fixes. Another CodeQL fix is in #2175. From the new issues, a couple are valid ones but most are FPs. I deliberately ignored note severity for now, feel free to enable those later (the workflow block can collapse to just `queries: +security-and-quality` then).

az0:
Thanks!

PR #2190, opened 2026-07-11
Validate data with real checks instead of assert

XhmikosR:
Asserts are stripped under python -O (and -OO in the frozen Windows build), so checks on untrusted input or safety invariants vanish in release builds.

This is potentially breaking and I only tested this on my Windows dev machine, but better fail loud that swallow the failure and cause bad side effects.

Please have a close look and let me know if you want me to make any changes.

### DIFF
--- a/bleachbit/CLI.py
+++ b/bleachbit/CLI.py
@@ -101,7 +101,8 @@ def args_to_operations_list(preset, all_but_warning):
     args = []
     if not backends:
         list(register_cleaners())
-    assert len(backends) > 1
+    if len(backends) == 0:
+        raise RuntimeError('no cleaners registered')
     for key in sorted(backends):
         c_id = backends[key].get_id()
         for (o_id, _o_name) in backends[key].get_options():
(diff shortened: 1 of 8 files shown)


bleachbit/CLI.py
```diff
@@ -101,7 +101,8 @@ def args_to_operations_list(preset, all_but_warning):
     args = []
     if not backends:
         list(register_cleaners())
-    assert len(backends) > 1
+    if len(backends) <= 1:
+        raise RuntimeError('no cleaners registered')
```
az0, line 105: What about this?
```py
if len(backends) == 0:
    raise RuntimeError('no cleaners registered')
```

bleachbit/WindowsWipe.py
az0:
```
while write_length_bytes > 0:
        assert write_length_bytes > 0
```

You didn't touch this code, but in case you want to change it, this assertion can be deleted

bleachbit/CLI.py
az0: Besides this, I didn't see any problems. 

Thanks!

XhmikosR: Done

PR #2191, opened 2026-07-12
Serve HTTP status codes from a local server in network tests

XhmikosR:
Should hopefully fix the random network test failures since we won't be hitting real sites. Not sure if it's the best solution, but IMHO it's crucial that CI is stable.

### DIFF
--- a/tests/TestNetwork.py
+++ b/tests/TestNetwork.py
@@ -9,10 +9,11 @@ Test case for module Network
 """
 
 # standard imports
+import http.server
 import ipaddress
 import logging
 import os
-import random
+import threading
 import unittest
 import warnings
 from unittest.mock import MagicMock, Mock, patch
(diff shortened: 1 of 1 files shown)


az0:
Yesterday I made commit 08b2278b9f2177aa9fe920756121734933b8c705 and today 046f10cbc303498ca3d06f1030375f7fa4a4959a that could help, so first I'll see how those work, but your solution also makes sense.

XhmikosR:
IMHO this is much preferable and cleaner. Only con we don't test https right now.

XhmikosR:
I personally keep hitting these errors, for example https://github.com/XhmikosR/bleachbit/actions/runs/29275856857/job/86904597631

XhmikosR:
I still think we should merge this. It doesn't make any sense for us to rely on external services and have CI failures for that reason, when we can achieve the same in a proper, robust way.

PR #2192, opened 2026-07-12
Switch to GitHub Actions CI

XhmikosR:
Some notes:

1. I'm not sure about the large file extent tests.
2. I dropped the SQLite loops to 5 because they increased the Windows CI time. We could make this only for Windows.
3. I went with fast build because it's significantly slower. I added back the English installer later as you can see.
4. ~~I also added a workflow to clean up CI caches so that they don't accumulate and hit the 1GB limit. This might not be needed in this repo, though so I can split that to another branch in case it's needed later.~~
5. **You will need to add the new secrets to the repo secrets. I would suggest that you delete them from AppVeyor after a few days and delete the repo from the integration. Assuming everything's fine with the switch :)**
6. **The build number will reset. Not sure how to handle this or if it's a real problem.**

I have tried to pin all deps to prevent supply chain attacks and to ensure the builds are consistent and controlled. IMHO everything should be in the repo pinned and handled by dependabot on a monthly basis which I added. Feel free to update/tweak the deps later if needed.

Note that GHA runs in parallel not sequentially like AppVeyor. Which means, we don't need to wait like before. Also, I didn't add `concurrency` on purpose but we could add it for the non-master branches later if needed. Personally I wouldn't care about total CI minutes for now. Although, it'd be nice to speed things up. Python 3.14 is super fast as it seems.


.github/workflows/cleanup-caches.yml
```diff
@@ -0,0 +1,28 @@
...
+  pull_request:
+    types:
+      - closed
+
+permissions: {}
+
+jobs:
+  cleanup:
+    name: Delete caches for the closed PR
+    runs-on: ubuntu-latest
```
az0: [ubuntu-slim](https://github.com/actions/runner-images/blob/main/images/ubuntu-slim/ubuntu-slim-Readme.md) could be enough

.github/workflows/cleanup-caches.yml
```diff
@@ -0,0 +1,28 @@
...
+      actions: write # delete the PR branch caches
+
+    steps:
+      - name: Delete caches
+        env:
+          GH_TOKEN: ${{ github.token }}
+          GH_REPO: ${{ github.repository }}
+          BRANCH: refs/pull/${{ github.event.pull_request.number }}/merge
+        run: |
+          gh cache list --ref "${BRANCH}" --limit 100 --json key --jq '.[].key' \
```
az0: I'm not sure we need cache cleanup, so I am okay to skip this for now.

- The [cache management page](https://github.com/bleachbit/bleachbit/actions/caches) shows 39 caches with sizes of ~6MB to ~84MB, most at ~11MB.
- [Free tier has 10GB](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#increasing-cache-size)
- I can pay for more cache expensive, if needed

---

The only resource that I've had an issue with on the GitHub free tier is Actions storage. Below is an email from a few months ago, and I resolved it by enabling billing. 

The PR here will significantly increase Action storage by uploading the Windows artifacts, but that is okay. It's just $0.25 USD/GB/month ([reference](https://docs.github.com/en/billing/concepts/product-billing/github-actions)).

> You have used 100% of the Actions storage included for the bleachbit account
> GitHub Actions Usage
 > Your plan includes 0.5 GB of Actions storage per month at no extra cost. You have used 100% so far this billing cycle.
> 0.5 GB used / 0.5 GB included
> You'll be billed for usage beyond the included amount - unless a $0 budget is set for Actions, which blocks further usage until it resets. Your usage will reset in 13 days on April 01, 2026.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+      - 'doc/**'
+      - 'docker/**'
+      - 'MANIFEST.in'
+      - 'org.bleachbit.BleachBit.desktop'
+      - 'org.bleachbit.BleachBit.metainfo.xml'
+      - 'org.bleachbit.policy'
+      - 'po/**'
+      - 'README.md'
+      - 'windows/**'
+      - 'windows/**'
```
az0: tests.yaml now tests and builds Windows, so it needs to trigger on changes to many files there. A commonly change file that could be excluded is `windows/NsisInclude/NsisMultiUserLang.nsh`: it's updated with translations.

`windows/nsis_translations.py` is another candidate to ignore.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+      - 'bleachbit.spec'
+      - 'COPYING'
+      - 'debian/**'
+      - 'doc/**'
+      - 'docker/**'
+      - 'MANIFEST.in'
+      - 'org.bleachbit.BleachBit.desktop'
+      - 'org.bleachbit.BleachBit.metainfo.xml'
+      - 'org.bleachbit.policy'
+      - 'po/**'
```
az0: In https://github.com/bleachbit/bleachbit/blob/master/.github/workflows/test_macos.yaml I handled it by flipping `paths-ignore` to `paths`, but now this looks simpler

```suggestion
      - 'po/*.po' # do not ignore Makefile
      - 'po/*.pot'
```

This is not a blocker

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+          $version = & $env:PYTHON_EXE -c "import bleachbit; print(bleachbit.APP_VERSION)"
+          $dest = "s3://bleachbitci/dl/$version.$($env:GITHUB_RUN_NUMBER)-$($env:GITHUB_REF_NAME)/"
+          Write-Host "Deploying to $dest"
+          Get-ChildItem BleachBit-*-setup.exe, BleachBit-*-setup-English.exe, BleachBit-*-portable.zip |
+            ForEach-Object { aws s3 cp $_.FullName $dest }
+
+  coveralls-finish:
+    needs: [test, windows]
+    if: github.repository == 'bleachbit/bleachbit'
+    runs-on: ubuntu-latest
```
az0, line 353:
```suggestion
    runs-on: ubuntu-slim
```

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+# 2025-05-01: ubuntu-latest is currently 24.04, and its native
+# python version is 3.12.
+# native python for 22.04 is 3.10.
+jobs:
+  test:
```
az0, line 47: To distinguish from windows (which tests also---plus builds exe/zip), please change to ubuntu or linux

```suggestion
  linux:
```

Lines below will need to match this

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,358 @@
...
+          "%PYTHON_EXE%" -m coverage combine --data-file=.coverage .coverage.admin .coverage.nonadmin .coverage.psexec
+          "%PYTHON_EXE%" -m coverage report --data-file=.coverage
+
+      - name: Generate lcov coverage
+        if: github.repository == 'bleachbit/bleachbit'
+        shell: cmd
+        run: '"%PYTHON_EXE%" -m coverage lcov -o coverage.lcov --data-file=.coverage'
+
+      - name: Upload coverage to Coveralls
+        if: github.repository == 'bleachbit/bleachbit'
```
az0, line 315: The current behavior is to upload to Coveralls, even for PRs from forks. It's not used often, but it could provide helpful feedback, especially with large edits to `bleachbit/*.py`.

tests/TestFileUtilities.py
```diff
@@ -1535,7 +1535,8 @@ def test_sqlite_loop(self):
         """
         warnings.simplefilter("error")
         start_time = time.time()
-        for _ in range(50):
+        reps = 5 if os.getenv('CI') else 50
```
az0: This hasn't been an issue, so you could change it unconditonally

.github/workflows/cleanup-caches.yml
az0: I'm not sure we need cache cleanup, so I am okay to skip this for now.

- The [cache management page](https://github.com/bleachbit/bleachbit/actions/caches) shows 39 caches with sizes of ~6MB to ~84MB, most at ~11MB.
- [Free tier has 10GB](https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#increasing-cache-size)
- I can pay for more cache expensive, if needed

---

The only resource that I've had an issue with on GHA  is storage. Below is an email from a few months ago, and I resolved it by enabling billing. 

> You have used 100% of the Actions storage included for the bleachbit account
> GitHub Actions Usage
 > Your plan includes 0.5 GB of Actions storage per month at no extra cost. You have used 100% so far this billing cycle.
> 0.5 GB used / 0.5 GB included
> You'll be billed for usage beyond the included amount - unless a $0 budget is set for Actions, which blocks further usage until it resets. Your usage will reset in 13 days on April 01, 2026.

tests/TestFileUtilities.py
```diff
@@ -1535,7 +1535,8 @@ def test_sqlite_loop(self):
...
+        for _ in range(reps):
```
az0: This has not been an issue for a while, so the count could be unconditionally lowered

```suggestion
        for _ in range(5):
```

windows/python-gtk3-install.ps1
```diff
@@ -54,18 +61,18 @@ if ($env:APPVEYOR -ne $true) {
...
 # Python and GTK+
 $GTK_ZIP_FN = "gtk3.24-x86-windows.zip"
 if (-not (Test-Path $GTK_ZIP_FN)) {
     Write-Host "Downloading Python and GTK+..."
     Invoke-WebRequest -Uri "$base_download_url/$GTK_ZIP_FN" -OutFile $GTK_ZIP_FN
-    Get-FileHash -Path $GTK_ZIP_FN -Algorithm SHA256 | Format-List
 } else {
     Write-Host "Python and GTK+ are already downloaded."
 }
+Assert-FileHash $GTK_ZIP_FN "3578518592813af187bd4244b5bf0ab4bb867f7654e993ff6e73466a222c74c8"
```
az0, line 75: Thanks!

.github/workflows/cleanup-caches.yml
XhmikosR: We should revisit storing artifacts then. Maybe drop the retention days from 90 to 60.

EDIT: Setting compression for artifacts to best might also help.
EDIT2: I switched to `9`

.github/workflows/tests.yaml
XhmikosR: I'll leave the new ignores for you to handle later. I removed the windows ignore for now.

.github/workflows/tests.yaml
XhmikosR: You seem to prefer slim but it doesn't really gain you a lot AFAICT.

.github/workflows/tests.yaml
XhmikosR: Yes, but then it fails for every fork job... It only works here.

.github/workflows/tests.yaml
XhmikosR: Better you edit the files on my fork since I don't know what exactly you prefer here. I can rebase later before the merge.

XhmikosR:
I think I addressed most of your comments except for a couple.

.github/workflows/tests.yaml
XhmikosR: As you can see coveralls still runs on PRs. It just skips direct runs on forks.

XhmikosR:
I think we should drop the ubuntu-slim changes. It's not worth the slowness. CodeQL for example takes 4-6 times more time. I personally don't want to spend time waiting for CI :)

az0:
> I think I addressed most of your comments except for a couple.

I will take a closer look---thank you.


> I think we should drop the ubuntu-slim changes. It's not worth the slowness. CodeQL for example takes 4-6 times more time. I personally don't want to spend time waiting for CI :)

I'm not hard against ubuntu-latest, but two thoughts

1. CodeQL isn't the slowest workflow, so the total time for workflows doesn't change. 
2. Here are some statistics from: [Action Performance Metrics: workflow_file_name:codeql.yml](https://github.com/bleachbit/bleachbit/actions/metrics/performance?tab=jobs&filters=workflow_file_name%3Acodeql.yml)

image

Analyze (python): 1m 4s vs 3m47s
Analyze (actions): 41s vs 1m38s

XhmikosR:
CodeQL can be super slow too. This was taking 20 minutes and i cancelled it on my fork: https://github.com/XhmikosR/bleachbit/actions/runs/29232797824/job/86760543389

Generally, you really don't gain a lot from ubuntu-slim.

That being said, Windows tests need to be parallelized, but that's for another time.

.github/workflows/cleanup-caches.yml
az0: The artifact windows-build-6.0.2.13.zip is a 60.MB zip file that contains two .exe and one .zip. The exe and inner zip are already compressed, so level 9 compression on the outside may just make it slower.

In another project involving GHA artifacts, I tried to avoid the nested zip, but there wasn't a way to avoid it, so fast compression may be optimal.

XhmikosR: Yeah, it doesn't look like it has any effect, but it's not noticeably slower. I will revert it, though, since it doesn't gain us anything.

And yes, there's no way currently to skip the nested zip, it's been like that since the beginning.

XhmikosR:
I think the only thing left is the workflow names? Just edit my branch and I will rebase. Just note the TODOs in my first post.

XhmikosR:
I pushed a couple of minor fixes. What do you think about https://github.com/XhmikosR/bleachbit/commit/8129b7da19db3f452c0a2acb257fc2cce7daa08b? Not sure I prefer trusting chocolatey... You are already trusting them for psexec, though. It's less YAML code, but comes with the trust issue. Your call.

az0:
Thank you.

The other day, I brought your changes into branch `windows_on_gha` and am working on them there before merging to the `master` branch.

XhmikosR:
Just make sure you pick up the latest changes. Let me know about the above patch if you want.

az0:
>What do you think about [XhmikosR@8129b7d](https://github.com/XhmikosR/bleachbit/commit/8129b7da19db3f452c0a2acb257fc2cce7daa08b)? Not sure I prefer trusting chocolatey... You are already trusting them for psexec, though. It's less YAML code, but comes with the trust issue. 

I like the diff is +4 -30, but (1) it already works as is and (2) despite chocolatey having a clean reputation, there have been published compromises in other repositories. Recently, it was AUR (Arch Linux).

IIRC, `choco` was not my first choice for psexec, but I ran into an obstacle.

I considered mirroring these dependencies. 

> Just make sure you pick up the latest changes. 

I grabbed your update_translations change. Thanks.

> Let me know about the above patch if you want.

The patch is appealing, but it's not worth the small risk. Thank you, though.

az0:
Thanks for all this. After 11 years, no more AppVeyor CI.

I merged your changes, minus the changes for Windows fast build mode. Instead, I split up Windows testing, building, and deployment.

Dependabot is now running.

I started working on pytest for Windows. It is faster: in the [last run](https://github.com/bleachbit/bleachbit/actions/runs/29473440985), standard unittest took 6m 46s vs 3m44s for pyttest. In the [run before](https://github.com/bleachbit/bleachbit/actions/runs/29471970664), 7m27s vs 3m23. However, I didn't merge this because of crashes, related in part to logging.

XhmikosR:
@az0: make sure you follow the other TODO from my OP, especially number 5. :)

az0:
Yes, thanks.

The new deployment upload is working, but for the files to show, I need to make backend changes. Then, CI files will show on download.bleachbit.org instead of ci.bleachbit.org, so everything is one place.

In AppVeyor, I disabled the repo integration

XhmikosR:
All right. BTW I didn't split the workflows on purpose, because the total time spent is more this way.

I still think the solution for Windows CI is one job, just making the tests parallel if possible. Also, updating to Python 3.14 should speed things up judging by the Linux CI time.

az0:
>All right. BTW I didn't split the workflows on purpose, because the total time spent is more this way.

Earlier you wrote, "I personally don't want to spend time waiting for CI," so I expected you would prefer it split up to reduce wall time.

>Also, updating to Python 3.14 should speed things up judging by the Linux CI time.

PyGObject and GTK do not work easily on Windows with Python, so BleachBit uses [Python3 from vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/python3/vcpkg.json) which is on 3.12.13, and it's been difficult to update 3.12.9 to 3.12.13. That's a WIP in dev branch of [pygtkwin repo](https://github.com/bleachbit/pygtkwin).

Updating it to Python 3.14 would be more difficult, but it will be easier for the TUI and wxWidgets front ends. The TUI front end now uses [vanilla Python 3.13 64-bit](https://github.com/bleachbit/bleachbit/blob/tui/appveyor.yml). So does the [wxWidgets front end](https://github.com/bleachbit/bleachbit/blob/wx-gui-mvp/appveyor.yml).

XhmikosR:
> Earlier you wrote, "I personally don't want to spend time waiting for CI," so I expected you would prefer it split up to reduce wall time.

Yes, but there's a balance to keep :) Now we run plenty of duplicate stuff and because GHA run in parallel it is faster but you are spending more total time. That being said, if we cannot parallelize tests on Windows, I guess the current way is a viable workaround.

> PyGObject and GTK do not work easily on Windows with Python, so BleachBit uses [Python3 from vcpkg](https://github.com/microsoft/vcpkg/blob/master/ports/python3/vcpkg.json) which is on 3.12.13, and it's been difficult to update 3.12.9 to 3.12.13. That's a WIP in dev branch of [pygtkwin repo](https://github.com/bleachbit/pygtkwin).
> 
> Updating it to Python 3.14 would be more difficult, but it will be easier for the TUI and wxWidgets front ends. The TUI front end now uses [vanilla Python 3.13 64-bit](https://github.com/bleachbit/bleachbit/blob/tui/appveyor.yml). So does the [wxWidgets front end](https://github.com/bleachbit/bleachbit/blob/wx-gui-mvp/appveyor.yml).

Yeah, I'm not aware of all the internals, I just notice that Python 3.14 is a lot faster. Actually, even 3.13 and probably 3.12 is faster too, just not the same. Basically anything below that looks like it's too slow.

PR #2194, opened 2026-07-12
Python 3 cleanup

XhmikosR:


az0:
Two changes are partial: do you want to complete them here before I merge? Or I could merge as is. Let me know.

- TestInit.py still uses unittest.makeSuite

- Python 2-style super() used in TestGUI, TestNetwork, etc.

XhmikosR:
Good catch, I missed those. Should hopefully be OK now. Only markovify should be left, but I think it's third-party code.

az0:
> Good catch, I missed those. Should hopefully be OK now. Only markovify should be left, but I think it's third-party code.

Yes, and I made a full replacement for markovify that is faster.

PR #2196, opened 2026-07-12
CleanerML: skip the unused world-writable stat on Windows

XhmikosR:


bleachbit/CleanerML.py
```diff
@@ -366,11 +366,11 @@ def list_cleanerml_files(local_only=False, system_only=False):
             cleanerdirs += (bleachbit.local_cleaners_dir, )
     if not local_only and bleachbit.system_cleaners_dir:
         cleanerdirs += (bleachbit.system_cleaners_dir, )
+    check_world_writable = sys.platform != 'win32'
```
az0: please

```suggestion
    check_world_writable = IS_WINDOWS
```

and then at the top: `from bleachbit import IS_WINDOWS`

XhmikosR: Shouldn't be the opposite? I pushed a new patch with the correct check.

az0: 😂

PR #2197, opened 2026-07-13
Drop executable bit from assets and menu.ui

XhmikosR:


XhmikosR:
Note that I didn't remove the executable bit from py files. Only a few of them need it AFAICT.

PR #2198, opened 2026-07-13
Tighten test_macos.yaml

XhmikosR:
Non-whitespace diff: https://github.com/bleachbit/bleachbit/pull/2198/changes?w=1

@az0: note that this was triggering a CodeQL error. Better pay attention to those from now on after ignoring the FPs in GitHub security tab.

### DIFF
--- a/.github/workflows/test_macos.yaml
+++ b/.github/workflows/test_macos.yaml
@@ -23,12 +23,16 @@ on:
       - 'po/Makefile'
       - '!README.md'
       - '!windows/**'
+  workflow_dispatch:
+
+permissions:
+  contents: read
 
 # Headless tests on macOS. GTK is not installed, so GUI tests
 # auto-skip via @unittest.skipUnless(HAVE_GTK, ...).
 jobs:
   test:
-    timeout-minutes: 25
+    timeout-minutes: 30
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/test_macos.yaml
```diff
@@ -38,45 +42,48 @@ jobs:
...
+        run: |
+          brew install gettext
+          echo "$(brew --prefix gettext)/bin" >> "$GITHUB_PATH"
 
-    - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v5
-      with:
-        python-version: ${{ matrix.python-version }}
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v5
```
az0: Do you want to bump setup-python to v6 and pin with SHA?

https://github.com/actions/setup-python

XhmikosR: I thought I did, but apparently I missed it. Should be OK now.

XhmikosR: BTW maybe you haven't enabled dependabot in the repo settings, because it should have already opened a PR.

EDIT: When you enable it make sure you enable group security updates too.

az0: Previously enabled (Settings - Advanced Security):

- dependency graph
- dependency alerts
- dependency security updates

Now enabled: grouped security updates

PR #2201, opened 2026-07-14
Cleaner fixes

XhmikosR:


cleaners/claude.xml
```diff
@@ -32,7 +32,7 @@ later.  See the COPYING file in the top-level directory.
     <description>Delete the logs</description>
     <action command="delete" search="walk.files" path="~/.claude/debug/"/>
     <action command="delete" search="walk.files" path="~/.claude/logs/"/>
-    <action command="delete" search="file" path="~/.claude/" regex="\.log$"/>
+    <action command="delete" search="glob" path="~/.claude/*.log" type="f"/>
```
XhmikosR, line 35: AFAICT this did nothing before. Not sure if there are any logs there, though since I don't have Claude installed myself.

az0: Good catch

This line was added by origin0722, so I wrote him a comment
https://github.com/bleachbit/bleachbit/pull/2144/changes#r3580221193

cleaners/vlc.xml
```diff
@@ -56,6 +56,6 @@
   <option id="memory_dump">
     <label>Memory dump</label>
     <description>Delete the file</description>
-    <action command="delete" search="file" path="$$Profile$$\crashdump"/>
+    <action command="delete" search="file" path="$$Profile$$/crashdump"/>
```
az0, line 59: Based on short research and quick test, I could not make vlc 3.0.23-2-0-g79128878dd to make this directory on Linux (Ubuntu).

```terminal
vlc -vv &
killall -SIGSEGV vlc
find ~/.config/vlc ~/.cache/vlc /var/lib/systemd/coredump/
```

I did confirm your change is needed for it to match by manually making it.

```
mkdir -p ~/.config/vlc/crashdump
touch ~/.config/vlc/crashdump/testing
```

So I could accept your change, or we could research more whether this is Windows specific. If it is Windows only, then `os="windows"` could be added at the option level.

cleaners/palemoon.xml
```diff
@@ -27,7 +27,7 @@
   <running type="exe" os="windows" same_user="true">palemoon.exe</running>
   <running type="pathname">~/.moonchild productions/pale moon/*/lock</running>
   <var name="base">
-    <value os="windows">%AppData%\moonchild productions\value</value>
+    <value os="windows">%AppData%\moonchild productions\pale moon</value>
```
az0, line 30: Good catch

cleaners/vlc.xml
XhmikosR: This might be Windows only, then, but I cannot tell you for sure...

PR #2214, opened 2026-07-17
Wipe fixes

XhmikosR:


bleachbit/Unix.py
```diff
@@ -617,7 +617,7 @@ def apt_autoremove():
 def apt_autoclean():
     """Run 'apt-get autoclean' and return the size (un-rounded, in bytes) of freed space"""
     try:
-        return run_cleaner_cmd('apt-get', ['autoclean'], r'^Del .*\[([\d.]+[a-zA-Z]{2})}]', ['^E: '])
+        return run_cleaner_cmd('apt-get', ['autoclean'], r'^Del .*\[([\d.]+[a-zA-Z]{2})\]', ['^E: '])
```
az0: This is a good catch, but it should expect a space between the number and the units like `[\d.]+ ?[a-zA-Z]{2}`

Below is output from with apt 3.2.0 on Ubuntu 26.04

```
Del libcurl3t64-gnutls 8.18.0-1ubuntu2.2 [417 kB]
Del libcurl4t64 8.18.0-1ubuntu2.2 [426 kB]
Del curl 8.18.0-1ubuntu2.2 [272 kB]
```

bleachbit/Wipe.py
```diff
@@ -258,15 +258,17 @@ def wipe_write():
                 _('There was at least one file on a file system that does not support advanced overwriting.'), UserWarning)
             f = wipe_write()
         else:
-            # The wipe succeed, so prepare to truncate.
-            f = open(path, 'wb')
+            # The wipe succeeded and already overwrote the file in place.
+            # Reopen only to truncate; opening 'wb' would zero the length.
+            f = open(path, 'wb') if truncate else None
```
az0: This is an improvement, but non-blocking issue: in case of non-cluster aligned file, it will leave it larger. For example, a 4097B file will be left at 8192B.

The only use of truncate=False is in Memory.py, which does not work on Windows, so options are

1. Refactor `wipe_write` helper into a function that `Memory.py` can call directly, and remove `truncate` parameter from `wipe_contents` function.
1. Save original size before modifying file, then set EOF using `truncate`
2. If truncate=True on Windows, then raise exception

bleachbit/Unix.py
XhmikosR: Fixed. Now it follows what `apt_autoremove` does.

bleachbit/Wipe.py
XhmikosR: I went with the first option, let me know.

az0: You probably saw this

https://github.com/bleachbit/bleachbit/actions/runs/29591409509/job/87921143860

```
ERROR: test_wipe_swap_linux (tests.TestMemory.MemoryTestCase.test_wipe_swap_linux)
Test for wipe_swap_linux() with mocks
  File "/Users/runner/work/bleachbit/bleachbit/tests/TestMemory.py", line 307, in test_wipe_swap_linux
    with mock.patch('bleachbit.Memory.wipe_contents'):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/mock.py", line 1467, in __enter__
    original, local = self.get_original()
                      ^^^^^^^^^^^^^^^^^^^
  File "/Library/Frameworks/Python.framework/Versions/3.12/lib/python3.12/unittest/mock.py", line 1437, in get_original
    raise AttributeError(
AttributeError: <module 'bleachbit.Memory' from '/Users/runner/work/bleachbit/bleachbit/bleachbit/Memory.py'> does not have the attribute 'wipe_contents'
```

XhmikosR: Yeah, I missed the replacements there: https://github.com/bleachbit/bleachbit/compare/52637264f7bb98ae6a039dccc37ffd5c264bfac5..639a70f22240d69a4446431bce5e48cde005511a

PR #2219, opened 2026-07-18
Streamline "IS_*" constants and drop unused imports

XhmikosR:


bleachbit/Options.py
```diff
@@ -30,15 +30,15 @@
...
-if 'nt' == os.name:
-    from win32file import GetLongPathName
-
 # local application imports
 import bleachbit
-from bleachbit import General
+from bleachbit import General, IS_WINDOWS
 from bleachbit.Language import get_text as _
 
+# third-party imports
```
XhmikosR, line 38: I moved this block in order to use `IS_WINDOWS`, but I can revert this if you prefer.

PR #2223, opened 2026-07-19
Further performance tweaks

XhmikosR:
The gains will be mostly visible to Winapp2 users, and on Windows, but POSIX should benefit too, just not so much.

### DIFF
--- a/bleachbit/FileUtilities.py
+++ b/bleachbit/FileUtilities.py
@@ -936,13 +936,16 @@ def uris_to_paths(file_uris):
 def whitelisted_posix(path, check_realpath=True, _followed_link=False):
     """Check whether this POSIX path is whitelisted"""
     from bleachbit.Options import options
+    keep_paths = options.get_whitelist_paths()
+    if not keep_paths:
+        return False
     if check_realpath and os.path.islink(path):
         # also check the link name
         if whitelisted_posix(path, False):
             return True
         # resolve symlink
         return whitelisted_posix(os.path.realpath(path), False, _followed_link=True)
-    for (keep_type, keep_path) in options.get_whitelist_paths():
+    for (keep_type, keep_path) in keep_paths:
         if keep_type == 'file':
             if path_equal(path, keep_path):
                 return True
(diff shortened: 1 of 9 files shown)


bleachbit/FileUtilities.py
```diff
@@ -55,6 +54,12 @@
     # pylint: disable=invalid-name
     pywinerror = WindowsError
 
+# DirEntry.is_junction() was added in Python 3.12. Below that, fall back to
+# bleachbit.Windows.is_junction(), which stats the path itself instead of
+# reusing the DirEntry's cached data.
+# TODO: drop this fallback once the minimum Python version is 3.12+
+_DIRENTRY_HAS_IS_JUNCTION = hasattr(os.DirEntry, 'is_junction')
```
az0, line 61: On Windows, we can set the minimum Python to 3.12: we use Python 3.12 now through https://github.com/bleachbit/pygtkwin  . It's rare for BleachBit users on Windows to not use the provided BleachBit binary.

XhmikosR: I know but I went with the safest approach.

bleachbit/FileUtilities.py
```diff
@@ -258,35 +263,68 @@ def bytes_to_human(bytes_i):
...
+
+def _scan_children(top, list_directories, pending_dirs):
+    """Yield files under `top`, recursing into real subdirectories.
+
+    Symlinks and, on Windows, junctions are not descended into. When
+    list_directories is set, every directory found (including those links)
+    is collected into pending_dirs to be emitted after its contents.
+
+    Using os.scandir directly lets us reuse each entry's cached type instead
+    of re-stat'ing every subdirectory to test for links, as os.walk required.
```
az0, line 281: A future optimization is to pass the type and size to code like `FileUtilities.delete` (which checks for directories, junctions, etc.) and commands in `Command.py` (uses getsize).

XhmikosR: Feel free to push to my branch, you already have access since the checkbox " Allow edits and access to secrets by maintainers " is enabled :)

PR #2229, opened 2026-07-20
Remove deprecated app menu

City-busz:
It has not been used for years. Use the hamburger menu also on Windows.

See: https://wiki.gnome.org/Design/Whiteboards/AppMenuMigration

### DIFF
--- a/bleachbit/GuiApplication.py
+++ b/bleachbit/GuiApplication.py
@@ -100,24 +100,11 @@ class Bleachbit(Gtk.Application):
     def build_app_menu(self):
         """Build the application menu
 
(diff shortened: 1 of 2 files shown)


az0:
With this change, Windows has no application menu, so no way to access the menu items there.

image

Without this change, Windows has an application menu

image

image

City-busz:
I updated the commit to use the hamburger menu also on Windows.

I'm asking this, because I can see both the app menu and the hamburger menu on Linux if I set the button layout to `appmenu:minimize,maximize,close`:
bleachbit-appmenu
bleachbit-hamburgermenu

XhmikosR:
This has been manually merged on master.

PR #2236, opened 2026-07-25
CI and tests speedup

XhmikosR:
This is the best I could come up with. It should save up to ~25% when everything is cached. Not sure about the storage limits we might hit.

What this means is that even with these patches, you only have about 10-12 runs until you hit the action 2000 minutes free monthly limit. I'm not sure how this could be further improved...

Some ideas:

* Python 3.9 not using parallel tests is important and takes a lot of time
* stop using upx altogether; it's not so important and the distributed archive/installer are already compressed well
* use zstd instead of zip for assets you download and have control of; zstd is a lot faster and compresses much better
* stop using MSYS, venvs etc. Use whatever the runner has available and use setup-python action when possible
* use 64-bit for the whole thing, but that's a big change. In theory, though, this should give up to 25% speedup alone for Windows
* see if we can reduce the time spent on some tests; this seems the most realistic, although I don't know if it's even possible

I also tested running the windows tests in two jobs: the individual jobs time did decrease but the total usage was the same.

PS. I think the actions code should be minimal and most stuff should be in scripts/makefile etc

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,7 +1,2 @@
 [run]
 source = bleachbit
-
-[paths]
-source =
-    bleachbit/
-    C:\projects\bleachbit\bleachbit\
--- a/.github/actions/generate-revision/action.yml
+++ b/.github/actions/generate-revision/action.yml
@@ -1,0 +1,20 @@
+name: Generate Revision.py
+description: Write bleachbit/Revision.py with the current commit and build number
+
+outputs:
+  revision:
+    description: Short (8-char) commit SHA written to Revision.py
+    value: ${{ steps.generate.outputs.revision }}
+
+runs:
+  using: composite
+  steps:
+    - id: generate
+      shell: bash
+      run: |
+        rev="$(git rev-parse --short=8 HEAD)"
+        {
+          echo "revision = \"$rev\""
+          echo "build_number = \"${{ github.run_number }}\""
+        } > bleachbit/Revision.py
+        echo "revision=$rev" >> "$GITHUB_OUTPUT"
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -336,14 +336,21 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Cache downloaded dependencies
+      - name: Cache the installed environment
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
         with:
+          # Cache the downloaded archives AND the unpacked+pip-installed tree, so
+          # a warm run skips both the slow Expand-Archive and the pip installs.
+          # No restore-keys: a partial-key hit would restore a tree with stale
+          # dependencies that the install script would then skip re-installing.
+          # Test jobs don't mutate the tree, so caching it here is safe (unlike
+          # windows-build, whose shrink/UPX step strips files in place).
           path: |
             gtk*-x86-windows.zip
             gtk-themes.zip
             pygobject-*-win32.whl
-          key: win-test-deps-${{ hashFiles('.github/workflows/build_and_test.yaml', 'windows/python-gtk3-deps.lst', 'windows/python-gtk3-install.ps1') }}
+            vcpkg_installed
+          key: win-test-env-${{ hashFiles('.github/workflows/build_and_test.yaml', 'windows/python-gtk3-deps.lst', 'windows/python-gtk3-install.ps1', 'requirements.txt', 'windows/requirements.txt', 'requirements-test.txt') }}
 
       - name: Cache pip packages
         uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6.1.0
@@ -510,7 +517,7 @@ jobs:
         run: |
           # Sanitize
           $ref = ($env:GITHUB_REF_NAME -replace '[^A-Za-z0-9._/-]','_')
-          $ver = ($env:VERSION        -replace '[^A-Za-z0-9._-]','_')
+          $ver = ($env:VERSION         -replace '[^A-Za-z0-9._-]','_')
           $run = $env:GITHUB_RUN_NUMBER
           $attempt = $env:GITHUB_RUN_ATTEMPT
           $sha = $env:GITHUB_SHA.Substring(0,8)
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -153,10 +153,10 @@ jobs:
           export $(dbus-launch)
           sudo Xvfb :1 -screen 0 800x600x24 &
           sleep 5
-          if [[ "${{ matrix.python-version }}" == "3.14" ]]; then
-            dbus-run-session -- bash -c '/usr/lib/notification-daemon/notification-daemon & sleep 2 && xvfb-run -a bash -c "PYTHONPATH=. pytest -n auto"'
-          else
+          if [[ "${{ matrix.python-version }}" == "3.9" ]]; then
             dbus-run-session -- bash -c '/usr/lib/notification-daemon/notification-daemon & sleep 2 && xvfb-run -a bash -c "make tests && make tests-with-sudo"'
+          else
+            dbus-run-session -- bash -c '/usr/lib/notification-daemon/notification-daemon & sleep 2 && xvfb-run -a bash -c "make -C cleaners tests && PYTHONWARNINGS=error PYTHONPATH=. pytest -n auto && python windows/nsis_translations.py --unittest && make tests-with-sudo"'
           fi
 
       - name: Generate lcov coverage
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -395,7 +395,7 @@ jobs:
           COVERAGE_FILE: .coverage.admin
         run: |
           set PATH=%MSYS_BIN%;C:\ProgramData\chocolatey\bin;C:\Windows\system32;C:\Windows;C:\Windows\system32\wbem;C:\Windows\SysWOW64\WindowsPowerShell\v1.0\
-          "%PYTHON_EXE%" -m coverage run -m tests.TestAll
+          "%PYTHON_EXE%" -m pytest -n auto --dist=loadgroup --cov=bleachbit --cov-report= tests
 
       - name: Run non-admin tests
         if: matrix.suite == 'nonadmin'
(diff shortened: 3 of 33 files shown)


az0:
Have you looked into this  the coverage drop? I assume coverage files were not combined.

https://coveralls.io/jobs/185158483

> coverage: 47.529% (-15.4%) from 62.966%

XhmikosR:
Yeah, I saw that but I'm not sure it's related to my changes. Maybe coveralls is confused because I used the same branch name or something?

Because these checks are OK:

```
[coverage/coveralls](https://coveralls.io/builds/80789277)
coverage/coveralls — Coverage decreased (-0.4%) to 76.24%
[coverage/coveralls](https://coveralls.io/builds/80789277)
coverage/coveralls — Coverage decreased (-0.4%) to 76.24%
```

which shows that coverage was correctly combined.

XhmikosR:
@az0: I'm looking at the coverage issue, it doesn't seem right, indeed.

BTW, what's your take about upx or the other points I mention in my OP?

Using upx doesn't look it's worth it... fast build is almost 50% faster. And the savings are small in the installer and zip file, with the risk of AV software and everything else that comes from using upx.

XhmikosR:
NVM, the speedup must be from using 4 threads instead of 2 on master...

I think coverage is OK now. It was silently skipping GTK tests because there were packages missing from Python 3.14 where I switched the coverage to run on. Would be nice if this was a more visible in the future...

I also have [this patch](https://github.com/XhmikosR/bleachbit/commit/1693daf280702651ba648589fe2735a961e71d5d), but last time I tested it, it was the same in terms of minutes used. I'd leave it out of this PR for now, and we test it separately.

az0:
> This is the best I could come up with. It should save up to ~25% when everything is cached. 

25% is great! Thank you!

> Not sure about the storage limits we might hit.

GHA storage is the cheapest part, while the macOS per-minute cost is a major cost driver. 

> What this means is that even with these patches, you only have about 10-12 runs until you hit the action 2000 minutes free monthly limit. I'm not sure how this could be further improved...

The good news is GitHub waives the cost of all minutes for CI builds on public repositories.

I can pay for some minutes over the 2000, but not unlimited.

> * stop using upx altogether; it's not so important and the distributed archive/installer are already compressed well

> [from other comment] Using upx doesn't look it's worth it... fast build is almost 50% faster. And the savings are small in the installer and zip file, with the risk of AV software and everything else that comes from using upx.

My first choice is keep UPX for public builds, but it would be okay to keep UPX only for tagged releases (which are also public): neither have a financial cost for CI runner minutes. 

Are you worried about the wall time or the runner cost?

In the last UPX on the master branch, it took 56.8 seconds to save ~31 MB. 

https://github.com/bleachbit/bleachbit/actions/runs/30210885192/job/89816784268
2026-07-26 16:41:11 - INFO - Reduced size of the dist directory by 21,294,592 B from 52,029,377 B to 30,734,785 B in 56.8 s

We could save the UPX issue to not delay merging this PR.

 > * Python 3.9 not using parallel tests is important and takes a lot of time

I'd be okay switching Python 3.9 to run parallel tests. It would need pytest 8.4.2, which has a CVE that isn't a practical risk in single-user, ephemeral VM.

> * use zstd instead of zip for assets you download and have control of; zstd is a lot faster and compresses much better

zstd where?

1. coverage file for coveralls-finish (actions/upload-artifact) 
2. Windows distributables for windows-deploy (actions/upload-artifact) 
3. GHA cache
4. gtk3.24-x86-windows.zip

AFAIK, 1-3 do not support changing compression methods.

> * stop using MSYS, venvs etc. Use whatever the runner has available and use setup-python action when possible
> * use 64-bit for the whole thing, but that's a big change. In theory, though, this should give up to 25% speedup alone for Windows

You can compare to the wx and tui branches, which use vanilla 64-bit Python.

https://github.com/bleachbit/bleachbit/actions/workflows/build_wx_windows.yaml

build and test took 3m 48s last time, but it's not in parallel. The slowest setup step was installing pip, which this PR solves.

with your PR the "install Python and GTK" took 40 seconds. At a glance, it may not have a warm cache.

https://github.com/bleachbit/bleachbit/actions/runs/30212290779/job/89820450820
 
> PS. I think the actions code should be minimal and most stuff should be in scripts/makefile etc

Yes, some progress has been made in that direction.


Easy ideas to reduce CI time or costs: 

- combine the windows-deploy and windows-build jobs to avoid double overhead costs of starting up and because GitHub rounds up minutes
- switch windows-deploy (if not combined with windows-build) to ubuntu-slim because the per-minute cost is much cheaper and the job requires few resources
- replace ubuntu-latest with ubuntu-slim when fewer resources are needed
- in private repos, disable macOS by default. Run it either on demand or if commit message contains a trigger keyword

Also, there are non-GitHub platforms like AppVeyor, which has Windows and macOS.

XhmikosR:
> My first choice is keep UPX for public builds, but it would be okay to keep UPX only for tagged releases (which are also public): neither have a financial cost for CI runner minutes.
> 
> Are you worried about the wall time or the runner cost?
> 
> In the last UPX on the master branch, it took 56.8 seconds to save ~31 MB.
> 
> https://github.com/bleachbit/bleachbit/actions/runs/30210885192/job/89816784268 2026-07-26 16:41:11 - INFO - Reduced size of the dist directory by 21,294,592 B from 52,029,377 B to 30,734,785 B in 56.8 s
> 
> We could save the UPX issue to not delay merging this PR.

But if something then breaks with UPX you'd only notice it after a release.

The size decrease is real, but IMHO we should only care about the distribution size, i.e. zip and installer. And there, the gains are much smaller IIRC. That along with the extra time, AV flagging things, potential startup time increase, it's not worth it in eyes, that is all :)

It's not a blocker for this PR, but something to consider for later.

> I'd be okay switching Python 3.9 to run parallel tests. It would need pytest 8.4.2, which has a CVE that isn't a practical risk in single-user, ephemeral VM.

Something for a follow-up PR then.

> zstd where?
> 
>     1. coverage file for coveralls-finish (actions/upload-artifact)
> 
>     2. Windows distributables for windows-deploy (actions/upload-artifact)
> 
>     3. GHA cache
> 
>     4. gtk3.24-x86-windows.zip
> 
> 
> AFAIK, 1-3 do not support changing compression methods.

Caching through actions already uses zstd. I was thinking about the zip files you download and that you maybe have control of. Otherwise, my bad if it doesn't apply at all.

> You can compare to the wx and tui branches, which use vanilla 64-bit Python.
> 
> https://github.com/bleachbit/bleachbit/actions/workflows/build_wx_windows.yaml
> 
> build and test took 3m 48s last time, but it's not in parallel. The slowest setup step was installing pip, which this PR solves.
> 
> with your PR the "install Python and GTK" took 40 seconds. At a glance, it may not have a warm cache.
> 
> https://github.com/bleachbit/bleachbit/actions/runs/30212290779/job/89820450820

Personally I wouldn't even update pip to the latest version at all. Why do you need that? The GitHub Python versions already update their bundled pip AFAICT. Maybe not immediately after a release, but still.

As for 64-bit I will need to look closer if multiple threads are used etc.


> Easy ideas to reduce CI time or costs:
> 
>     * combine the windows-deploy and windows-build jobs to avoid double overhead costs of starting up and because GitHub rounds up minutes
> 
>     * switch windows-deploy (if not combined with windows-build) to ubuntu-slim because the per-minute cost is much cheaper and the job requires few resources
> 
>     * replace ubuntu-latest with ubuntu-slim when fewer resources are needed
> 
>     * in private repos, disable macOS by default. Run it either on demand or if commit message contains a trigger keyword

Sounds good, as long as it's not super slow.

XhmikosR:
Forgot to say, that, assuming you follow the PR approach for your branches too, running CI blindly on all branch pushes isn't ideal and results in duplicate runs, as you may have seen from the other repo.

XhmikosR:
One issue I notice but it's present on master too, is that I no longer see the menu or the icon on Windows 11. Unsure when this broke...

EDIT: must be a2b1a6da. Not sure how that landed on master since the PR is open.

az0:
> > My first choice is keep UPX for public builds, but it would be okay to keep UPX only for tagged releases (which are also public): neither have a financial cost for CI runner minutes.
> > Are you worried about the wall time or the runner cost?
> > In the last UPX on the master branch, it took 56.8 seconds to save ~31 MB.
> > https://github.com/bleachbit/bleachbit/actions/runs/30210885192/job/89816784268 2026-07-26 16:41:11 - INFO - Reduced size of the dist directory by 21,294,592 B from 52,029,377 B to 30,734,785 B in 56.8 s
> > We could save the UPX issue to not delay merging this PR.
> 
> But if something then breaks with UPX you'd only notice it after a release.
> 
> The size decrease is real, but IMHO we should only care about the distribution size, i.e. zip and installer. And there, the gains are much smaller IIRC. That along with the extra time, AV flagging things, potential startup time increase, it's not worth it in eyes, that is all :)

[Discussion 1817](https://github.com/bleachbit/bleachbit/discussions/1817) is a complaint about the increase in size from version 4.6.2 when UPX was disabled for the 5.0.0 stable release.

A few years ago, it was common to have one or two false positives on VirusTotal for each release, but lately it's improved.
 

> > I'd be okay switching Python 3.9 to run parallel tests. It would need pytest 8.4.2, which has a CVE that isn't a practical risk in single-user, ephemeral VM.
> 
> Something for a follow-up PR then.

After merging this PR, here's the [action run timing](https://github.com/bleachbit/bleachbit/actions/runs/30218647484):

linux 3.9: 3m50s rounds up to 4 minutes
linux 3.12: 1m11s rounds up to 2 minutes
linux 3.14 1m22s rounds up to 2 minutes

The two minute savings is $0.006 x 2 = $0.012, which is modest.

An idea is to change python 3.9 to `ubuntu-22.04-arm` and keep it on unittest:

- More diverse coverage because of CPU and unittest
- ARM costs $0.005 ($0.001 less than x64)

Either way, there's bigger opportunities than investing in this issue.

> Caching through actions already uses zstd. I was thinking about the zip files you download and that you maybe have control of. Otherwise, my bad if it doesn't apply at all.

in bleachbit/pygtkwin, `gtk3.24-x86-windows.zip` it would be okay to maximize compression, especially for releases

> Personally I wouldn't even update pip to the latest version at all. Why do you need that? The GitHub Python versions already update their bundled pip AFAICT. Maybe not immediately after a release, but still.

I'm okay to skip updating pip

> Forgot to say, that, assuming you follow the PR approach for your branches too, running CI blindly on all branch pushes isn't ideal and results in duplicate runs, as you may have seen from the other repo.

Your proposal is to run CI on PRs and not on push?

> One issue I notice but it's present on master too, is that I no longer see the menu or the icon on Windows 11. Unsure when this broke...

> EDIT: must be https://github.com/bleachbit/bleachbit/commit/a2b1a6da4ad45d2370609c961d79e9e25775de96. Not sure how that landed on master since the PR is open.

OK, I filed an issue https://github.com/bleachbit/bleachbit/issues/2246 and later noticed you fixed it here (PR #2245). Thanks

PR #2237, opened 2026-07-25
Windows setup: pass build commands as argument lists

XhmikosR:


windows/setup.py
```diff
@@ -182,13 +185,12 @@ def sign_files(filenames):
     Passing multiple filenames in one function call can be faster than
     two calls.
     """
-    filenames_str = ' '.join(filenames)
     if os.path.exists('CodeSign.bat'):
-        logger.info('Signing code: %s', filenames_str)
-        cmd = f'CodeSign.bat {filenames_str}'
+        logger.info('Signing code: %s', ' '.join(filenames))
+        cmd = ['CodeSign.bat'] + filenames
```
az0: `filenames` is tuple, so to avoid TypeError, I suggest

```suggestion
cmd = ['CodeSign.bat', *filenames]
```

az0:
I made a note about a TypeError, but otherwise, this is good. Thank you!

This change is a helpful robustness and security improvement for handling quoting commands during the Windows build. It would avoid potentially running something like `c:\program.exe`.

XhmikosR:
Thanks, I'll fix it later :)

az0:
I noticed that there is no check of exit codes for 7z, UPX, strip, and NSIS. There are indirect checks: 7-zip, strip, and NSIS check for the output file. UPX does inline replacement, so it would catch an error only by `assert_execute_console`.

This was preexisting, and it's not blocking the merge of the PR. You are welcome to add that, or I could add it later.

XhmikosR:
@az0: please check now

PR #2244, opened 2026-07-26
Drop EOL whitespace and convert tab to spaces

XhmikosR:


XhmikosR:
The .spec file is untested. The rest should be OK.

PR #2245, opened 2026-07-26
Fix missing menu on Windows

### LINKED ISSUE #2246: application icon missing on Windows
One issue I notice but it's present on master too, is that I no longer see the menu or the icon on Windows 11. Unsure when this broke...

EDIT: must be a2b1a6da. Not sure how that landed on master since the PR is open.

_Originally posted by @XhmikosR in https://github.com/bleachbit/bleachbit/issues/2236#issuecomment-5085159471_

XhmikosR:
Regression from a2b1a6da.

I can only test on Windows, unsure about other platforms. BTW a2b1a6da is merged on master but #2229 is not. Another solution would be to revert the offending patch.

Fixes #2246

### DIFF
--- a/bleachbit/GuiApplication.py
+++ b/bleachbit/GuiApplication.py
@@ -96,12 +96,9 @@ class Bleachbit(Gtk.Application):
         return application_id_suffix
 
     def build_app_menu(self):
-        """Build the application menu
+        """Register the actions used by the application menu
 
-        On Linux with GTK 3.24, this code is necessary but not sufficient for
-        the menu to work. The headerbar code is also needed.
-
-        On Windows with GTK 3.18, this code is sufficient for the menu to work.
+        The menu itself is built by the headerbar code in GuiWindow.
         """
         from bleachbit.Language import setup_translation
         setup_translation()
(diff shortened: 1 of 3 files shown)


az0:
in the last stable release (and many past releases), the menu was accessible via application on the left of title bar, and the PR moves it to the hamburger on right of title bar, making it consistent with how Linux has already been. 

Consistency with past releases would also be good, but we could try it this way. Consistency between platforms (Windows/Linux) is good too, though for a new user on Windows, either option can be weird. 

There is a startup hint "Access the application menu by clicking the logo on the title bar" that will need corrected to use the same hint as for Windows "Access the application menu by clicking the hamburger icon on the title bar." However, I will merge this PR now to "stop the bleeding."

Screenshot of the PR (e9c1b9748ebccaa3a5c584667f452cb094555656) on Ubuntu

image

With fix on Windows 11

image

The broken state without the fix (c8dc496956bf153f385042de8d429bdc7425c32f) on WIndows 11

image

Old screenshot of BleachBit 5.1.1 with application menu on left



Even though the icon is no longer shown, the .ico and .png icons are still needed.

PR #2258, opened 2026-08-04
Fix CLI KeyError on -cleaner.option negation

### LINKED ISSUE #2028: CLI Negation fails with KeyError Traceback when excluding cleaners
Describe the bug
The CLI fails with a KeyError and a Python Traceback when attempting to use the negation operator (-) to exclude a specific cleaner.

To Reproduce

Open a command prompt/terminal.

Navigate to the BleachBit installation directory.

Run the following command (or similar):
bleachbit_console.exe --preview -- system.* -system.recycle_bin

Observe the terminal output.

The error happens every time.

Expected behavior
The CLI should process the system.* argument to include all system cleaners, but then interpret the -system.recycle_bin argument as a directive to exclude the Recycle Bin from that set, as per the CLI documentation.

Screenshots

Image

Desktop:
OS: Windows 11

BleachBit version: 5.10

Additional context
The traceback indicates the failure happens in Worker.py within the run_operations method. Specifically:

KeyError: The application is looking for a cleaner literally named -system.recycle_bin instead of recognizing the minus sign as a negation toggle.

Permission Issues: The log also shows Permission denied when writing to bleachbit.ini, suggesting the CLI may need to be run with Administrator privileges, though this is secondary to the parsing crash.

JokeGbenro:
## Summary

Fixes #2028. Excluding a cleaner from the command line with the `-` negation prefix (e.g. `system.* -system.recycle_bin`) crashed with a `KeyError` instead of excluding the option.

The argument `-system.recycle_bin` was being parsed as a cleaner literally named `-system` with option `recycle_bin`, producing an operations dict like `{'-system': ['recycle_bin'], ...}`. That bogus key later reached `Worker.run_operations`, where `backends['-system']` raised `KeyError` and dumped a traceback.

## Reproduction (before)

```
bleachbit_console.exe --preview -- system.* -system.recycle_bin
```

Intent: clean all system options *except* the Recycle Bin. Result: `KeyError` traceback, no cleaning.

## Fix

In `args_to_operations` (`bleachbit/CLI.py`), arguments beginning with `-` are now stripped of the prefix and routed into the same exclusion path already used by `--except`. So `system.*` is expanded first, then `-system.recycle_bin` removes the Recycle Bin option. If a negation removes a cleaner's last remaining option, the cleaner key is dropped entirely rather than left as an empty/stale entry.

The caller's `excludes` list is copied before appending so negations don't mutate it.

## Tests

Added cases to `test_args_to_operations` in `tests/TestCLI.py`:
- `-cleaner.option` negation removes the option
- Leading-`-` negation combined with an explicit `--except`
- A negation that removes a cleaner's last option drops the cleaner key (guards against the `KeyError`)

All `args_to_operations` / `--except` tests pass. The only failing test in the suite, `test_gui_exit`, fails identically on the base branch and is unrelated (GUI/environment).

## Notes

- Docstring for `args_to_operations` updated to document the negation syntax.
- On the command line, `-cleaner.option` args must follow `--` (as in the issue's example) so optparse treats them as positional arguments rather than options.

### DIFF
--- a/bleachbit/CLI.py
+++ b/bleachbit/CLI.py
@@ -130,7 +130,9 @@ def args_to_operations(args, preset, all_but_warning, excludes=None):
     """Convert command-line arguments to a dictionary of operations
 
     Args:
-        args: List of cleaner.option strings (e.g., ['system.tmp', 'firefox.cache'])
+        args: List of cleaner.option strings (e.g., ['system.tmp', 'firefox.cache']).
+            An argument prefixed with '-' (e.g., '-system.recycle_bin') is treated
+            as an exclusion, the same as passing it via --except.
         preset: Boolean indicating whether to use saved preset operations
         all_but_warning: Boolean indicating whether to include all non-warning operations
         excludes: Optional list of cleaner.option strings to remove from operations
@@ -146,8 +148,25 @@ def args_to_operations(args, preset, all_but_warning, excludes=None):
 
     if excludes is None:
         excludes = []
+    else:
+        # Copy so the leading-'-' negations below are not appended to the
+        # caller's list.
+        excludes = list(excludes)
+
+    # Support the leading '-' negation syntax (e.g. 'system.*'
+    # '-system.recycle_bin') by routing negated arguments into the same
+    # exclusion path as --except.  Without this, '-system.recycle_bin' was
+    # treated as a cleaner literally named '-system', which later raised a
+    # KeyError in Worker.run_operations.
+    included_args = []
+    for arg in args:
+        if arg.startswith('-') and len(arg) > 1:
+            excludes.append(arg[1:])
+        else:
+            included_args.append(arg)
+
     positive_args = set(
-        args + args_to_operations_list(preset, all_but_warning))
+        included_args + args_to_operations_list(preset, all_but_warning))
 
     def fix_deprecated(cid, oid):
         if 'system' == cid and 'free_disk_space' == oid:
--- a/tests/TestCLI.py
+++ b/tests/TestCLI.py
@@ -94,7 +94,15 @@ class CLITestCase(common.BleachbitTestCase):
              {'adobe_reader': ['cache', 'mru', 'tmp']}),
             (['adobe_reader.mru'], [], {'adobe_reader': ['mru']}),
             (['adobe_reader.*'], ['adobe_reader.cache'],
-             {'adobe_reader': ['mru', 'tmp']}))
+             {'adobe_reader': ['mru', 'tmp']}),
+            (['adobe_reader.*', '-adobe_reader.cache'], [],
+             {'adobe_reader': ['mru', 'tmp']}),
+            # Leading '-' negation combined with an explicit --except (issue #2028)
+            (['adobe_reader.*', '-adobe_reader.mru'], ['adobe_reader.cache'],
+             {'adobe_reader': ['tmp']}),
+            # A negation that removes every option drops the cleaner entirely,
+            # rather than leaving a '-adobe_reader' key that later raises KeyError
+            (['adobe_reader.mru', '-adobe_reader.mru'], [], {}))
         for test_args, excludes, expected in tests:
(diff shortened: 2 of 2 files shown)


JokeGbenro:
Hi @az0 pls can you check this out also

MohammedAlkindi (commented):
Ran this on Windows 11 / Python 3.13.13.

**No new failures.** `tests/TestCLI.py` on this branch vs a clean `upstream/master` checkout on the same machine:

```
this branch : 1 failed, 32 passed, 1 skipped, 2 subtests passed
master      : 1 failed, 32 passed, 1 skipped, 2 subtests passed
```

The single failure is `test_gui_exit` (`AssertionError: 1 != 0`) on both, which is a GUI-dependent test on a headless-ish Windows box, not something this PR touches.

**The added cases have teeth.** Reverting only `bleachbit/CLI.py` and keeping the new table entries:

```
$ git checkout upstream/master -- bleachbit/CLI.py
$ python -m pytest tests/TestCLI.py::CLITestCase::test_args_to_operations -q
E   AssertionError: {'-adobe_reader': ['cache'], 'adobe_reader': ['cache', 'mru', 'tmp']}
                 != {'adobe_reader': ['mru', 'tmp']}
FAILED tests/TestCLI.py::CLITestCase::test_args_to_operations
```

and restored, `1 passed`. That failure output is the bug itself — the bogus `'-adobe_reader'` key is exactly what reaches `backends[...]` and raises the `KeyError` in #2028, so the assertion pins the cause rather than a symptom downstream of it.

One thing I could not check: the `bleachbit_console.exe --preview` reproduction from your PR description, since that needs an installed build rather than a source checkout. The parsing-level coverage above is from `args_to_operations` directly, which is where the malformed key originates.

JokeGbenro:
Thanks for validating this. I’ve addressed the review by keeping the fix at the parsing source in `args_to_operations`, where the malformed `'-adobe_reader'` / `'-system'` key originates.

I also added one more focused regression check to confirm the caller-provided `excludes` list is copied and not mutated when shorthand negations are appended internally. This preserves existing `--except` semantics while covering the `-cleaner.option` path.

I re-ran the targeted tests:

- `tests/TestCLI.py::CLITestCase::test_args_to_operations` passes
- `tests/TestCLI.py::CLITestCase::test_parse_cmd_line_except` passes

The broader `tests/TestCLI.py` run still shows only the known pre-existing Windows `test_gui_exit` failure (`AssertionError: 1 != 0`), matching your earlier result and unrelated to this CLI parsing fix.

MohammedAlkindi:
Re-ran the same checks against the updated head (`63ac214`), native Windows 11, Python 3.13.13:

- `tests/TestCLI.py::CLITestCase::test_args_to_operations` and `::test_parse_cmd_line_except` — both pass, including the new negation rows and the assertion that the caller's `excludes` list comes back unmutated.
- The cause is still pinned at the new location: with this head's tests and only `bleachbit/CLI.py` reverted to master, `test_args_to_operations` fails with `{'-adobe_reader': ['cache'], 'adobe_reader': ['cache', 'mru', 'tmp']} != {'adobe_reader': ['mru', 'tmp']}` — the malformed key that produces the reported `KeyError`. Restoring this PR's `CLI.py` returns it to green.
- Full `tests/TestCLI.py`: `1 failed, 32 passed, 1 skipped, 2 subtests passed` — the one failure is the pre-existing Windows `test_gui_exit` (`AssertionError: 1 != 0`), identical to master's baseline on this machine and unrelated to this change.

Same result as before, at the better location. Looks ready from my side.

JokeGbenro:
Thanks @MohammedAlkindi for revalidating the updated changes and confirming the regression coverage.

@az0, the targeted tests are passing, and the remaining `test_gui_exit` failure has been confirmed to be pre-existing and unrelated to this PR. The latest review indicates the fix is ready. When you have a chance, could you please take a final look then merge?

I’m happy to address any additional feedback.

PR #2263, opened 2026-08-11
fix: parse apt sizes smaller than 10,000 bytes

### LINKED ISSUE #2262: Failure to parse apt output if < 10000 bytes
BleachBit's regex fails in the less common scenario of output smaller than 10,000 bytes. Example outputs from apt: `Del foo 1.0-1 [1234 B]` or `After this operation, 1234 B disk space will be freed.`

Notes

- This issue is a follow up to 0c9d1b7e41fb7fe2487700724bb46b589c18880d regarding parsing output of `apt`.
- apt's logic is in function `SizeToStr`in `strutl.cc`
- apt uses base 10 (not base 2)
- apt suffixes are 'k' (lowercase) and then 'M' through 'Y' (uppercase)
- We do not support Z or Y, which is not a real problem (especially in _this economy_ for storage hardware).
- We already set LC_ALL=C, so don't need to worry about localization for numbers.
- `apt` vs `apt-get` produce different strings
- Adding `-s` flag to `apt get autoremove` to simulate (dry run) skips the message, so it is not helpful.

## Normal usage (no bug)

Real-world procedure to reproduce output for autoremove with size >10000B (common):

```
sudo apt install -y mc
sudo apt remove -y mc
echo n | sudo apt autoremove # show without removing
sudo python3 bleachbit.py -c apt.autoremove
```

Example output from apt 3.2.0 on Ubuntu 26.04

```
$ sudo apt autoremove 
REMOVING:                       
  mailcap  mc-data

Summary:
  Upgrading: 0, Installing: 0, Removing: 2, Not Upgrading: 22
  Freed space: 6,521 kB

Continue? [Y/n] 
(Reading database… 463673 files and directories currently installed.)
Removing mailcap (3.75ubuntu1.1)…
Removing mc-data (3:4.8.33-1.1build1)…
Processing triggers for man-db (2.13.1-1build1)…
Processing triggers for hicolor-icon-theme (0.18-2build1)…
```

## Bug

```
sudo apt-get install -y ttf-staypuft # install small package
sudo apt-mark auto ttf-staypuft
echo n | sudo LC_ALL=C apt-get autoremove
```

Above outputs "After this operation, 9216 B disk space will be freed."

Then BleachBit reports 0B

```
$ sudo python3 bleachbit.py -c apt.autoremove
apt-get autoremove 0B 
Disk space recovered: 0B
Files deleted: 0
Special operations: 1

$ dpkg -l ttf-staypuft
┌── Desired=Unknown/Install/Remove/Purge/Hold
│┌─ Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
││┌ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
│││ Name           Version      Architecture Description
├┼┼─══════════════─════════════─════════════─═════════════════════════════════
un  ttf-staypuft   <none>       <none>       (no description available)

```

MohammedAlkindi:
Closes https://github.com/bleachbit/bleachbit/issues/2262

## Cause

Both apt regexes in `Unix.py` required exactly two letters for the unit, via `[a-zA-Z]{2}`. apt omits the prefix entirely below 10,000 bytes, so those lines never matched:

| line | old pattern | new pattern |
|---|---|---|
| `After this operation, 1234 B disk space will be freed.` | no match | `1234 B` |
| `After this operation, 44.0 kB disk space will be freed.` | `44.0 kB` | `44.0 kB` |
| `After this operation, 74.7MB disk space will be freed.` | `74.7MB` | `74.7MB` |
| `Del foo 1.0-1 [1234 B]` | no match | `1234 B` |
| `Del foo 1.0-1 [44.0 kB]` | `44.0 kB` | `44.0 kB` |

A non-matching line is skipped rather than raising, so the freed space was silently under-counted by the size of every package small enough to be reported in plain bytes.

## Why `[kMGTE]?B` and not `{1,2}`

`{1,2}` is the smaller edit and fixes the reported case, but it keeps the pattern accepting units the converter cannot handle. `human_to_bytes` supports `kMGTE` only, and the current `{2}` already matches `1.0 ZB`, which then reaches the converter and raises `ValueError`. Pinning the pattern to exactly the set `human_to_bytes` converts fixes the report and turns that latent exception into the same silent skip every other unrecognised line gets. That matches your note on the issue that Z and Y are not worth supporting.

Happy to switch to plain `{1,2}` if you would rather keep the diff literal to the report.

No change was needed in `human_to_bytes` — its own pattern already makes both the prefix and the `B` optional, so `human_to_bytes('1234 B')` returns `1234`.

## Tests

`tests/TestUnix.py` gains the bare-`B` case to the existing `test_apt_autoclean_mock`, and a new `test_apt_autoremove_mock` covering all three unit shapes. Both follow the mock pattern already used in that file, so they run on any platform.

They fail on `master` and pass with this change:

```
$ git stash push -- bleachbit/Unix.py          # revert only the fix
$ python -m pytest tests/TestUnix.py -k apt -q
>           self.assertEqual(apt_autoclean(), 1234)
E           AssertionError: 0 != 1234
>               self.assertEqual(apt_autoremove(), expected)
E               AssertionError: 0 != 1234
```

`0 != 1234` is the under-count itself: the line is skipped and `freed_space` never accumulates.

With the fix restored:

```
$ python -m pytest tests/TestUnix.py -q
27 passed, 33 skipped in 0.25s        # clean master: 26 passed, 33 skipped
$ python -m pytest tests/TestFileUtilities.py -q
54 passed, 12 skipped, 54 subtests passed
```

## What I did not run

Windows 11, Python 3.13.13, so I could not exercise apt end to end — the 33 skips above are the `@common.skipIfWindows` cases, unchanged in count from master. The parsing is covered by the mocked strings taken verbatim from the issue, and I have not claimed an integration run I did not do. I also deliberately did not run the full suite, since several cleaner tests touch real user paths and this is a live machine rather than a CI runner; I ran the file I changed plus `TestFileUtilities.py`, which owns `human_to_bytes`.


XhmikosR:
This needs a rebase.

MohammedAlkindi:
Rebased onto current master. The conflict was in Unix.py — upstream switched the same line to `General.resolve_exe('apt-get')`; kept that together with this PR's regex change. TestUnix on the rebased branch: 27 passed / 33 skipped, vs 26 / 33 on baseline master — the +1 is this PR's new test. CI on the PR is the authority for the rest.

bleachbit/Unix.py
```diff
@@ -614,7 +614,8 @@ def apt_autoremove():
     args = ['--yes', 'autoremove']
     # After this operation, 74.7MB disk space will be freed.
     # After this operation, 44.0 kB disk space will be freed.
-    freed_space_regex = r'.*, ([\d.]+ ?[a-zA-Z]{2}) disk space will be freed.'
+    # After this operation, 1234 B disk space will be freed.
+    freed_space_regex = r'.*, ([\d.]+ ?[kMGTE]?B) disk space will be freed.'
```
az0, line 618:
```suggestion
    freed_space_regex = r'.*, ([\d.]+ ?[kMGTPE]?B) disk space will be freed.'
```

bleachbit/Unix.py
```diff
@@ -625,7 +626,8 @@ def apt_autoremove():
 def apt_autoclean():
     """Run 'apt-get autoclean' and return the size (un-rounded, in bytes) of freed space"""
     try:
-        return run_cleaner_cmd(General.resolve_exe('apt-get'), ['autoclean'], r'^Del .*\[([\d.]+ ?[a-zA-Z]{2})\]', ['^E: '])
+        # Del foo 1.0-1 [1234 B]
+        return run_cleaner_cmd(General.resolve_exe('apt-get'), ['autoclean'], r'^Del .*\[([\d.]+ ?[kMGTE]?B)\]', ['^E: '])
```
az0, line 630: Add P to both regexes

```suggestion
        return run_cleaner_cmd(General.resolve_exe('apt-get'), ['autoclean'], r'^Del .*\[([\d.]+ ?[kMGTPE]?B)\]', ['^E: '])
```

PR #2268, opened 2026-08-17
Compress images

XhmikosR:
I think the Windows 10 theme are 3rd-party, but they are unoptimized. As for the BleachBit PNG file, it can be compressed to 8-bit and save a lot of bytes.

Your call if this should land @az0. Let me know if you want me to make any changes.


az0:
> I think the Windows 10 theme are 3rd-party

Yes

>  As for the BleachBit PNG file, it can be compressed to 8-bit and save a lot of bytes.

At a glance, it looks similar. 

> Your call if this should land. 

I've optimized png files before, and more optimization is appreciated.  Soon I'll take a closer look at this PR.

> Let me know if you want me to make any changes.

https://github.com/bleachbit/bleachbit-misc/blob/master/logo_svg_to_ico.sh is a script that converts .svg to .ico via .png, and it includes a png optimization step. Perhaps you would modify that repo to include your png optimization method for the .ico and more generally to all .png files in the main repo? 

BTW, if you are interested in more optimization, an idea is to look for more files in the py2exe frozen build (including inside library.zip) that can be removed. One method is to grab a CI portable build, delete some files, and check if it starts: this can be scripted.

XhmikosR:
> https://github.com/bleachbit/bleachbit-misc/blob/master/logo_svg_to_ico.sh is a script that converts .svg to .ico via .png, and it includes a png optimization step. Perhaps you would modify that repo to include your png optimization method for the .ico and more generally to all .png files in the main repo?

I'll try to have a look but this PR was simpler for me :)

I used pngquant to convert the files to 8-bit and then oxipng and zopflipng.

PR #2269, opened 2026-08-17
Further CodeQL fixes

XhmikosR:


.github/workflows/build_and_test.yaml
```diff
@@ -154,7 +154,7 @@ jobs:
           sudo Xvfb :1 -screen 0 800x600x24 &
           sleep 5
           if [[ "${{ matrix.python-version }}" == "3.9" ]]; then
-            dbus-run-session -- bash -c '/usr/lib/notification-daemon/notification-daemon & sleep 2 && xvfb-run -a bash -c "make tests"'
+            dbus-run-session -- bash -c '/usr/lib/notification-daemon/notification-daemon & sleep 2 && xvfb-run -a bash -c "make -C po tests && PYTHONWARNINGS=error PYTHONPATH=. pytest -n logical --dist=loadgroup -m '"'"'not no_xdist'"'"' && PYTHONWARNINGS=error PYTHONPATH=. pytest -m no_xdist && python windows/nsis_translations.py --unittest"'
```
az0, line 157: @XhmikosR 
Please add `$(MAKE) -C cleaners tests` or I can do it after merging.

Also, I am considering changing `make tests` (in project root directory) to use pytest instead of unittest, so code like this could use it.

bleachbit/Windows.py
```diff
@@ -1025,7 +1025,8 @@ def is_junction(path):
         if tag is not None:
             return tag == IO_REPARSE_TAG_MOUNT_POINT
     except (OSError, AttributeError):
-        pass
+        logger.debug('no reparse tag for %s, so falling back to '
```
az0, line 1028: @XhmikosR 

This broke `tests/TestCLI.py::CLITestCase::test_debug_log ` on Windows 

```
>               self.assertEqual(1, stderr.count('Traceback'))
E               AssertionError: 1 != 2

tests\TestCLI.py:174: AssertionError
```

XhmikosR: I worked around it, but I don't get why CI didn't fail... I didn't notice it at all.

XhmikosR:
The remaining CodeQL issues should be manually ignored in the GitHub UI. I think most, if not all, are false issues.

.github/workflows/build_and_test.yaml
XhmikosR: IIRC I intentionally skipped some stuff in the past, but if you want to streamline the test commands, feel free to adapt it. You can also commit and push directly on PR branches from forks.

PR #2272, opened 2026-08-19
Use os.path.isjunction, since os.is_junction never existed

XhmikosR:


bleachbit/Windows.py
```diff
@@ -1013,13 +1013,15 @@ def get_windows_version():
...
+    https://docs.python.org/3/library/os.path.html#os.path.isjunction
     """
     if not IS_WINDOWS:
         return False
-    if hasattr(os, 'is_junction'):
-        return os.is_junction(path)
+    # os.path.isjunction() was added in Python 3.12. Below that, read the
+    # reparse tag ourselves.
+    # TODO: drop this fallback once the minimum Python version is 3.12+
+    if hasattr(os.path, 'isjunction'):
```
az0, line 1023: @XhmikosR 
You may change `bleachbit/__init__.py` to require Python 3.12 for Windows (while still supporting older Python for POSIX systems).

https://github.com/bleachbit/bleachbit/blob/ca4855172e199ccbfb1bb640c619ff81f5bed646/bleachbit/__init__.py#L25-L27

XhmikosR: I'd rather you handle that later in a separate PR :)

Generally, the whole Python support situation is weird right now. I understand why, so personally, I'd rather use the same everywhere regardless of OS.

PR #2275, opened 2026-08-19
Fix winapp2.ini RegKey entries being ignored

### LINKED ISSUE #2270: 6.03 erroring on winapp2.ini regkey deletions
**Describe the bug**
Upon installing 6.03 and launching it I get the error " ignoring 'winreg' action from untrusted winapp2.ini: section=xxxxxxxxx " on every winapp2.ini entry that has a regkey deletion action.

**To Reproduce**
Launch BleachBit with a winapp2.ini file present

The error happens (sometimes/every time).  Every time.

**Expected behavior**
BleachBit should process the regkey deletions without error

**Screenshots**
Image

 - OS: Windows 11 x64 via python
 - BleachBit version 6.0.3 beta

**Additional context**

XhmikosR:
Fixes #2270

### DIFF
--- a/bleachbit/CleanerML.py
+++ b/bleachbit/CleanerML.py
@@ -431,19 +431,22 @@ def list_cleanerml_files(local_only=False, system_only=False):
 
 
 def is_trusted_cleaner(pathname):
-    """Return True if the cleaner file is in the system cleaners directory.
+    """Return True if the cleaner file sits next to the application.
 
     The personal cleaners directory is user-writable, so a lower-integrity
-    process could plant a cleaner there. In portable mode the system
-    directory is writable too, but so is the executable beside it, so
-    distrusting it would gain nothing.
+    process could plant a cleaner there. The system and local cleaners
+    directories are writable in portable mode and in the source tree, but so
+    is the executable beside them, so distrusting them would gain nothing.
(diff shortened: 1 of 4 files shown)


bleachbit/CleanerML.py
```diff
@@ -44,8 +44,11 @@
 
 logger = logging.getLogger(__name__)
 
-"""Actions blocked for untrusted (user-writable) cleaners."""
-UNTRUSTED_BLOCKED_COMMANDS = ('process', 'winreg')
+"""Actions blocked for untrusted (user-writable) cleaners.
+
+Deletion stays allowed, so blocking registry deletion too would buy nothing.
+"""
+UNTRUSTED_BLOCKED_COMMANDS = ('process',)
```
XhmikosR, line 51: @az0: this was my original approach, only blocking processes. I went with the tighter approach but it breaks the experience so I guess it's a compromise...

XhmikosR: If you have any better ideas, please let me know.

XhmikosR:
@az0: CodeQL timed out, might be worth increasing the timeout or just switching back to ubuntu-slim.

az0:
Thank you for doing this. On my first pass, this looked okay. I will take another look soon.

PR #2278, opened 2026-08-20
nsis_translations.py: write NsisMultiUserLang.nsh with CRLF

XhmikosR:
Fixes a warning on CI

Makes me wonder if we really need CRLF for those, though. I'll test it with LF too, **do not merge yet**.


XhmikosR:
It seems LF works and we don't need CRLF. Let me know and I'll squash the patches, or you can choose squash when merging the PR.

PR #2280, opened 2026-08-20
Run autopep8

XhmikosR:
All these tools, pylint, flake, autopep8, should have a dedicated CI run otherwise it's a little moot since they are not enforced...

### DIFF
--- a/bleachbit/Memory.py
+++ b/bleachbit/Memory.py
@@ -127,9 +127,10 @@ def make_self_oom_target_linux(uid=None):
         if uid is None:
             uid = General.get_real_uid()
         if uid > 0:
-            # TRANSLATORS: Debug message when a process gives up root/admin privileges.
-            # %(pid)d is the integer process ID; %(uid)d is the integer user ID to switch to.
-            drop_msg = _("Dropping privileges of process ID %(pid)d to user ID %(uid)d.")
+            drop_msg = _(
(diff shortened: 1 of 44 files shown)


bleachbit/Memory.py
```diff
@@ -129,7 +129,8 @@ def make_self_oom_target_linux(uid=None):
         if uid > 0:
             # TRANSLATORS: Debug message when a process gives up root/admin privileges.
             # %(pid)d is the integer process ID; %(uid)d is the integer user ID to switch to.
-            drop_msg = _("Dropping privileges of process ID %(pid)d to user ID %(uid)d.")
+            drop_msg = _(
```
az0, line 130: @XhmikosR 

After this change, gettext fails to match the translation hint to the string.

There's some documentation about this in `docs/CONTRIBUTING.md`.

A test is to run

```
make -C po bleachbit.pot`
grep -B5 "Dropping priv" bleachbit.pot
```

Which should show

```
#. TRANSLATORS: Debug message when a process gives up root/admin privileges.
#. %(pid)d is the integer process ID; %(uid)d is the integer user ID to switch to.
#: ../bleachbit/Memory.py
#, python-format
msgid "Dropping privileges of process ID %(pid)d to user ID %(uid)d."
```

az0:
Besides the loss of one translation hint, LGTM. Thank you

PR #2281, opened 2026-08-20
Makefile: run pyflake and pylint in parallel

XhmikosR:
This should cut the time down to a few seconds from minutes, at least on my dev Windows VM.

It introduces some new issues that pylint can catch now, though. I'll leave those to you @az0. Maybe disable some for now and tackle them later, your call.


az0:
Yes, you were right these tools were slow.

Both tools write normal findings to stdout, but pyflakes3 writes syntax errors to stderr, so how about `2>&1` to redirect all to the log?  Otherwise, LGTM

PR #2283, opened 2026-08-20
Declare charset_normalizer as an explicit dependency

### LINKED ISSUE #2113: evaluate charset-normalizer vs chardet
Inslide `library.zip`, the `chardet` directory is 971,323 bytes

### LINKED ISSUE #2276: chardet is huge in library.zip
After unpacking library.zip, the `chardet` folder is 12,111,039 B. 

The dependency is used only for `clean_ini`, which we use only in `vlc.xml`.

Potential options include

- Replace chardet it with charset_normalizer, which requests requires but is blocked in excluded in `windows/setup.py`.
- Check VLC's source code for which encodings it uses for its configuration file. It may always use the same encoding, so we don't need to detect it.

chardet-6.0.0.post1
BleachBit 6.0.3 beta (9e1654fb)

XhmikosR:
Didn't notice you had assigned #2276 to you @az0.

Anyway, feel free to skip it for 6.0.4 or drop it altogether.

Fixes #2276, closes #2113

### DIFF
--- a/bleachbit.spec
+++ b/bleachbit.spec
@@ -68,6 +68,7 @@ BuildArch:      noarch
 %if %{is_redhat_family}
 BuildRequires:  desktop-file-utils
 BuildRequires:  gettext
+BuildRequires:  python3-charset-normalizer
 BuildRequires:  python3-psutil
 BuildRequires:  python3-requests
 BuildRequires:  python3-setuptools
@@ -75,6 +76,7 @@ Requires(post): desktop-file-utils
 Requires(postun): desktop-file-utils
 Requires:       python3
 Requires:       gtk3
+Requires:       python3-charset-normalizer
 Requires:       python3-gobject
 Requires:       python3-psutil
 Requires:       python3-requests
@@ -92,6 +94,7 @@ BuildRequires:  %{pyprefix}-base
 %if 0%{?suse_version} == 1600
 BuildRequires:  python3-base
 %endif
+BuildRequires:  %{pyprefix}-charset-normalizer
 BuildRequires:  %{pyprefix}-psutil
 BuildRequires:  %{pyprefix}-requests
 BuildRequires:  %{pyprefix}-setuptools
@@ -102,6 +105,7 @@ BuildRequires:  %{pyprefix}-xml
 BuildRequires:  update-desktop-files
 Requires:       gobject-introspection
 Requires:       %{pyprefix}
+Requires:       %{pyprefix}-charset-normalizer
 Requires:       %{pyprefix}-gobject
 Requires:       %{pyprefix}-gobject-Gdk
 Requires:       %{pyprefix}-psutil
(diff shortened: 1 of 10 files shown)


az0:
Thank you for working on this issue. However, I took a different direction in 498abfbbb12a6b67865a7c3149c05123f17bd0c4 by dropping chardet without replacing it with any package. If you see a problem, let me know or submit a PR.

XhmikosR:
If it fixes the issue, that's fine by me :) BTW my solution just used a package that's already a dependency.

az0:
FYI: Now _requests_ throws a warning, though it seems harmless, and it could be suppressed.

```
requests\__init__.pyc:92: RequestsDependencyWarning: Unable to find acceptable character detection dependency (chardet or charset_normalizer).
```

BleachBit version = 6.0.3.235
Git revision = 53cc9131
gi.version = 3.55.0
GTK version = 3.24.51

XhmikosR:
Yeah, but that's your patch not the patch from this branch AFAICT. :)

az0:
Yes, you are totally right. I did not  mean to imply otherwise. I believe warning is harmless, but if not this PR could fix it

XhmikosR:
Your call. My solution seems to be less destructive with the same result since `charset_normalizer` is already a dependency of requests so it does look like a cleaner solution.

XhmikosR:
@az0: I rebased the branch and I noticed you only include charset_normalizer in windows requirements. My original patch included it everywhere.

az0:
> I rebased the branch and I noticed you only include charset_normalizer in windows requirements. My original patch included it everywhere.

Yes. 

_You are familiar with this story, but I'm summing it up to find the miscommunication. Let me know if I missed something._

Before 498abfbbb12a6b67865a7c3149c05123f17bd0c4, charset_normalizer was already excluded from the py2exe build, and the commit also excluded chardet also, which caused a warning from the requests package: it requires either package for encoding detection. Rather than suppress the warning in requests, your solution in 25df86a312ef4ef6a212cafad77e57a72720348d was to have py2exe package charset_normalizer.

On Windows, charset_normalizer provides encoding detection for Winapp.

Previously, chardet was used for encoding detection when in `clean_ini`, but since .ini cleaning is used only for VLC, which always uses UTF-8, now the `clean_ini` function detects only the BOM (which VLC always writes) vs no BOM: `clean_ini` no longer does full-blown encoding detection. The UTF-8 assumption avoids a dependency on either chardet or charset_normalizer, and it avoids extra CPU cycles and potential encoding misclassifications for VLC .ini files.

The cross-platform role for charset_normalizer is to prevent the warning from requests, and it's sufficient to to rely on transitive inclusion of charset_normalizer through normal Python dependency management (e.g., pip, apt, dnf).

In the future, the requests package may change its dependencies on charset_normalizer, so a hard-coded dependency here (like in the RPM .spec file) would be stale.

XhmikosR:
But IMHO it's better be explicit and drop the dep if it ever changes upstream than silently relying on it being present...

PR #2284, opened 2026-08-21
Use the language selected in the installer

### LINKED ISSUE #2031: Installer ignores manual language selection and defaults to "Auto-detect"
Describe the bug
When running the Windows installer in a specific language (e.g., Portuguese), the application does not persist this choice. Upon the first launch, the "Languages: Automatic" preference is enabled by default, overriding the user's installation intent.

To Reproduce
1. Run the BleachBit Windows installer.
2. Select a language other than English (e.g., Portuguese) from the setup dropdown.
3. Complete the installation and launch BleachBit.
4. Navigate to Edit > Preferences.
4. Observe that the Languages: Automatic checkbox is ticked, and the interface has reverted to the system default (English).

The error happens every time.

Expected behavior
If a user explicitly selects a language during the installation process, that language should be written to the configuration, and the "Auto-detect" feature should be disabled to respect the user's manual choice.

Screenshots

Desktop 
OS: Windows 11
BleachBit version: 5.1.0

Additional context
This creates a significant barrier for non-English speakers. If the installer asks for a language, the application must honor that selection rather than forcing the user to find the preferences menu in a language they may not understand to fix it.

XhmikosR:
Fixes #2031

**UNTESTED, please confirm if it fixes the issue.**

### DIFF
--- a/bleachbit/Language.py
+++ b/bleachbit/Language.py
@@ -20,10 +20,17 @@
 import gettext
 import os
 import logging
+import re
 
 from bleachbit import IS_POSIX, IS_WINDOWS
 
 logger = logging.getLogger(__name__)
+
+# Written by the Windows installer next to the executable, so a per-machine
+# install seeds every account on the machine.
+INSTALLER_LANGUAGE_FILENAME = 'installer-language.txt'
+
+_LANGUAGE_CODE_RE = re.compile(r'\A[a-z]{2,3}(_[A-Z]{2})?\Z')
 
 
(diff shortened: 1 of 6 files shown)


XhmikosR:
I'm going to close this since I don't really use the installer myself. Feel free to pick up the branch later, if you think it's worth it.

az0:
If it's okay with you, I'll leave it open for the next feature release. I can be responsible for any changes

PR #2290, opened 2026-08-23
Tolerate files vanishing from %temp% in TestWinapp preview

XhmikosR:
Trying to fix https://github.com/bleachbit/bleachbit/actions/runs/32604193414/job/97106750189

### DIFF
--- a/tests/TestCleaner.py
+++ b/tests/TestCleaner.py
@@ -19,7 +19,6 @@ import bleachbit
 from bleachbit import IS_WINDOWS, IS_POSIX
 from bleachbit.Action import ActionProvider, Command
 from bleachbit.Cleaner import Cleaner, backends, create_simple_cleaner, simpler_cleaner_process_path, register_cleaners
-from bleachbit.PathUtils import path_startswith
 
 from tests import common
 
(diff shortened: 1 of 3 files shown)


az0:
I started on that here, so I can compare them

https://github.com/bleachbit/bleachbit/commit/51c8ff0c1d4d2a89c5024ade2da3c6c8c87726b7
https://github.com/bleachbit/bleachbit/commit/59f6ea17c6ada713fdf46ed32de3337ca4f12034
https://github.com/bleachbit/bleachbit/commit/c32a5f1da48818416a4f3af1eb06e67658b86c4f

XhmikosR:
Yeah, ideally tests shouldn't depend on anything any other test does.

az0:
These are my final versions on master now
https://github.com/bleachbit/bleachbit/commit/cbebfde8286e4c3d76e850fb3d6e5e7a45ddeff2
https://github.com/bleachbit/bleachbit/commit/cbebfde8286e4c3d76e850fb3d6e5e7a45ddeff2

Your approach had several advantages, such as the use of `os.path.realpath`. If you want to look into combining our approaches, we could keep this open, or we could move on.

More importantly, it would be helpful to have a canonical method for path comparison to use for this (volatile test directories), ProtectedPaths.py, and the keep list (whitelisting) system with robust support for case sensitivity, extended path prefixes (Windows), noramlization, etc.

XhmikosR:
I'm going to close this for now, then. It'd be easier to base your patches on the existent ones instead of now having to resolve all the conflicts :)

PR #2297, opened 2026-08-26
Optimize DeepScan path construction

JoshSantiagoKC:
## Summary

DeepScan repeatedly calls os.path.join(dirpath, filename) while scanning files in the same directory. This repeats path-processing work even though dirpath is constant for every file in that directory. This change constructs a platform-correct directory prefix once per walked directory using:

```python
os.path.join(dirpath, '')
```

Filenames are then appended to that prefix in the hot scan loop.

Existing two-argument `CompiledSearch.match()` callers remain compatible. Search ordering, regex matching, result ordering, and command behavior are unchanged.

## Performance

Benchmarked locally on Windows with Python 3.12.6 using a deterministic synthetic tree of 50,000 files.

The benchmark itself is not included in this PR.

| Workload | Before median | After median |
| --- | ---: | ---: |
| 100 dirs × 500 files, 1 rule | 0.0909s | 0.0835s |
| 100 dirs × 500 files, 5 rules | 0.1460s | 0.1474s |
| 100 dirs × 500 files, 10 rules | 0.3668s | 0.2521s |
| 1,000 dirs × 50 files, 10 rules | 1.4683s | 0.5588s |
| 10 dirs × 5,000 files, 10 rules | 0.4110s | 0.4160s |

The improvement is workload-dependent. The representative 50,000-file / 10-rule workload improved by approximately 31%, while lower-rule and few-directory workloads ranged from neutral to modestly faster.

Profiling the 50,000-file / 10-rule workload showed the intended reduction in path-processing work:

- `os.path.join`: approximately 101,300 calls → 401 calls
- Regex evaluations: 500,000 → 500,000

The final candidate measured a 0.2582s median on the representative 50,000-file / 10-rule workload.

### Linux

The same workloads were re-run on Linux (Python 3.12.3, `posixpath.join`) against `master` vs this branch. Trees were sibling directories of empty files. Two of the ten rules always constructed a full path; the remaining rules exited on a filename regex. That is the mix that produced the Windows ~101k → 401 `os.path.join` counts.

| Workload | Before median | After median | Change |
| --- | ---: | ---: | ---: |
| 100 dirs × 500 files, 1 rule | 0.0493s | 0.0324s | −34% |
| 100 dirs × 500 files, 5 rules | 0.0949s | 0.0622s | −34% |
| 100 dirs × 500 files, 10 rules | 0.1128s | 0.0792s | −30% |
| 1,000 dirs × 50 files, 10 rules | 0.1232s | 0.0915s | −26% |
| 10 dirs × 5,000 files, 10 rules | 0.1183s | 0.0816s | −31% |

Absolute times are lower on Linux because `posixpath.join` is cheaper than `ntpath.join`. The representative 50,000-file / 10-rule improvement is consistent with Windows (~30% vs ~31%). On Linux the gain was consistent across all five shapes; Windows had been neutral on the 1-rule and 10×5,000 cases.

Profiling the 50,000-file / 10-rule workload:

- `os.path.join`: 100,300 calls → 401 calls
- Regex evaluations: 500,000 → 500,000

### macOS

Re-run on macOS 26.5.2, Apple M4 Max, APFS, Python 3.14.3, `master` (`cf8156e8`) vs this branch (`9b1aa26d`). Separate worktrees, alternated within each round (15 samples per variant). Rules are the 12 `cleaners/deepscan.xml` actions that apply on macOS, not synthetic regexes.

| Workload | Before median | After median | Change |
| --- | ---: | ---: | ---: |
| 100 dirs × 500 files, 1 rule | 0.0550s | 0.0485s | −11.9% |
| 100 dirs × 500 files, 5 rules | 0.1104s | 0.0961s | −13.0% |
| 100 dirs × 500 files, 10 rules | 0.1942s | 0.1664s | −14.3% |
| 1,000 dirs × 50 files, 10 rules | 0.2091s | 0.1812s | −13.3% |
| 10 dirs × 5,000 files, 10 rules | 0.1903s | 0.1650s | −13.3% |
| 10,000 dirs × 3 files, 12 rules | 0.3043s | 0.2817s | −7.4% |

Before/after ranges did not overlap: the slowest after sample was faster than the fastest before sample in every row.

macOS improved on every shape, including the 1-rule and 10×5,000 cases that were neutral on Windows. The difference appears primarily driven by rule composition. wholeregex-only rules require full-path construction for every file, while filename-regex rules can return before constructing a path. `CompiledSearch.match()` returns early on a filename-regex miss, so a `wholeregex`-only rule (`angular`, `node_modules`, `venv`, `pycache`) joins on every file. The first five real actions include two such rules, which predicts ~100,000 joins and matches the Windows profile (~101,300). Workloads whose rules all have a filename regex have no joins to remove; an 8-rule regex-only control measured −1.0% / −1.5% (noise, no regression).

Profiling the 50,000-file / 10-rule workload:

- `posixpath.join`: 205,360 calls → 401 calls (same 401 as Windows and Linux)
- Total function calls: −54%

Correctness: SHA-256 of sorted match lists agreed for all 12 tree×rule combinations. macOS also covers a case Windows cannot: `normalized_walk` recomposes filenames to NFC but leaves `dirpath` as stored, so a decomposed directory produces a mixed NFD-dir / NFC-file path. On a forced-NFD tree both branches return identical output, and every path still resolves on disk. `/`, `//`, trailing-slash, and doubled-separator cases also agree.

These benches are warm-cache. A cold-cache scan is more I/O-bound, so the end-to-end share of the win will be smaller.

## Tests

- `py -m unittest tests.TestDeepScan`: 12 passed, 1 skipped
- Added regression coverage for path-prefix construction, including filesystem root paths
- `git diff --check`: passes
- Full local suite: 707 tests, 6 failures, 2 host-dependent errors, 192 skips
- No DeepScan-related failures were observed in the full suite

The remaining full-suite failures are missing local dependencies, not Windows-specific behavior. On macOS they were 5× missing `psutil` plus `setuptools` for the Makefile test; installing both made the suite pass on both branches. That also covers `TestFileUtilities` paths that exercise `is_normal_directory` / `whitelisted`, which DeepScan calls.

### Linux

- `python3 -m unittest tests.TestDeepScan`: 12 passed, 2 skipped (Darwin walk + Windows junction)
- Path-prefix regression, including filesystem root `/`, passed
- `git diff --check`: passes
- Full local suite: 707 tests, 4 host-dependent failures, 134 skips
- No DeepScan-related failures were observed in the full suite

### macOS

- `python3 -m unittest tests.TestDeepScan`: 12 passed, 1 skipped (matches Windows)
- Path-prefix regression has teeth: mutating the prefix to `dirpath + os.sep` fails with `'//foo.txt' != '/foo.txt'`
- Full suite run on both branches: no new failures, no DeepScan-related failures

### DIFF
--- a/bleachbit/DeepScan.py
+++ b/bleachbit/DeepScan.py
@@ -73,14 +73,17 @@ class CompiledSearch:
         self.wholeregex = re_compile(search.wholeregex)
         self.nwholeregex = re_compile(search.nwholeregex)
 
-    def match(self, dirpath, filename):
+    def match(self, dirpath, filename, path_prefix=None):
         if self.regex and not self.regex.search(filename):
             return None
 
         if self.nregex and self.nregex.search(filename):
             return None
 
-        full_path = os.path.join(dirpath, filename)
+        if path_prefix is None:
+            full_path = os.path.join(dirpath, filename)
+        else:
+            full_path = path_prefix + filename
 
         if self.wholeregex and not self.wholeregex.search(full_path):
             return None
@@ -121,10 +124,11 @@ class DeepScan:
                     if not whitelisted(os.path.join(dirpath, dirname))
                     and is_normal_directory(os.path.join(dirpath, dirname))
                 ]
+                path_prefix = os.path.join(dirpath, '')
                 for c in compiled_searches:
                     # fixme, don't match filename twice
                     for filename in filenames:
-                        full_name = c.match(dirpath, filename)
+                        full_name = c.match(dirpath, filename, path_prefix)
                         if full_name is None:
                             continue
                         # fixme: support other commands
--- a/tests/TestDeepScan.py
+++ b/tests/TestDeepScan.py
@@ -84,6 +84,22 @@ class DeepScanTestCase(common.BleachbitTestCase, WindowsLinksMixIn):
                 continue
             self.assertLExists(cmd.path)
 
+    def test_compiled_search_path_prefix(self):
+        """CompiledSearch.match should preserve root path semantics."""
+        from bleachbit.DeepScan import CompiledSearch
+
+        compiled = CompiledSearch(
+            Search(command='delete', regex=r'\.txt$')
+        )
+
+        for dirpath in (self.tempdir, os.path.abspath(os.sep)):
+            with self.subTest(dirpath=dirpath):
+                path_prefix = os.path.join(dirpath, '')
+                self.assertEqual(
+                    compiled.match(dirpath, 'foo.txt', path_prefix),
+                    os.path.join(dirpath, 'foo.txt'),
+                )
+
     def test_skip_whitelisted_directory(self):
         """DeepScan should not search whitelisted directories."""
         # Reminder: each test case gets a new options directory,


XhmikosR:
You need to rebase against the current master branch and see if this still applies.

EDIT: actually you are up to date. @az0 do you know why AppVeyor fired here?

az0:
> do you know why AppVeyor fired here?

AppVeyor was partially disabled, so it didn't make sense that it tried on this PR and not other PRs. However, I did more thoroughly disable it now.

az0:
@JoshSantiagoKC 

This looks good, thanks! The blocking issue is a merge conflict. Could you take a look?

On an Apple M1, it was 28% faster (50k files, 5 rules including two wholeregex).  On Windows VM (AMD Ryzen 7 4800U), it was 36-37% faster (50K files, 10 mixed rules). Both reflect warm cache scenarios. All returned same (functional) results.

A non-blocking issue is handling of a bare UNC root like `\\server\share`. This is not accessible under normal circumstances with the current version, but in the future, it could be helpful to let users set the roots for each option (like `venv` cache).

PR #2315, opened 2026-09-01
Update action.yml

XhmikosR:
Pass explicitly run_number through env


az0:
I see, this is GitHub's recommended pattern: [Use an intermediate environment variable](https://docs.github.com/en/actions/reference/security/secure-use#use-an-intermediate-environment-variable). Thanks!

Issue #2042, opened 2026-03-01
Bleach bit fails to launch as root


tmick2:
**Describe the bug**
Bleachbit won't launch as root, it takes the root password then nothing. It launches as a standard user. 

**To Reproduce**
1.Launch Bleachbit as root from GUI
2. Input root password 
3. launch fails
4. Open CLI and become root
5. try to launch bleachbit 
6. Receive this:  bleachbit --preset -p

(bleachbit:100999): Gdk-CRITICAL **: 15:32:36.068: gdk_screen_get_root_window: assertion 'GDK_IS_SCREEN (screen)' failed
No work to do. Specify options. 
If I run 
bleachbit --debug-log=BLEACH_BUG
Debug log file initialized at BLEACH_BUG
running cmd loginctl
running cmd loginctl show-session 2 -p Type

(bleachbit:101274): Gdk-CRITICAL **: 15:35:06.891: gdk_screen_get_root_window: assertion 'GDK_IS_SCREEN (screen)' failed
user_locale: en_US, locale_dir: /usr/share/locale/
running cmd locale -a
GTK settings not found
BleachBit version = 5.0.2
Git revision = efb5cfb
gi.version = 3.54.5
SQLite version = 3.46.1
local_cleaners_dir = None
locale_dir = /usr/share/locale/
options_dir = /root/.config/bleachbit
personal_cleaners_dir = /root/.config/bleachbit/cleaners
system_cleaners_dir = /usr/share/bleachbit/cleaners
locale.getlocale = ('en_US', 'UTF-8')
os.getenv(DESKTOP_SESSION) = None
os.getenv(LOGNAME) = root
os.getenv(USER) = root
os.getenv(SUDO_UID) = None
os.path.expanduser(~") = /root
get_distribution_name_version() = debian n/a
sys.argv = ['/usr/bin/bleachbit', '--debug-log=BLEACH_BUG']
sys.executable = /usr/bin/python3
sys.version = 3.13.12 (main, Feb  4 2026, 15:06:39) [GCC 15.2.0]
__file__ = /usr/share/bleachbit/bleachbit/SystemInformation.py

The error happens every time.

**Expected behavior**
to launch and run Bleachbit ass root to clean files. 

**Screenshots**
N/A

**Desktop (please complete the following information):**
 - OS: Debian Forky Mate 1.28.2
 - BleachBit version 5.0.2

**Additional context**
N/A

az0:
[...]

PR #2316, opened 2026-09-02
Use pkexec instead of consolehelper on Red Hat

XhmikosR:
I only tested this by directly editing my desktop file on Fedora 44. Unsure if other OS'es need a similar change.

Refs #2042. It probably closes that issue too assuming Debian-based repos already work, but I can't test there.

### DIFF
--- a/bleachbit.spec
+++ b/bleachbit.spec
@@ -49,13 +49,6 @@
 %define has_fdupes 1
 %endif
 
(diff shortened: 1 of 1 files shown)


az0:
Thank you for working on this. I'd like to test more carefully, so I marked it for 6.1.0.

Issue #2298, opened 2026-08-27
BleachBit 6.0.3 — macOS Tahoe Apple Silicon Build & GTK Quartz Support + dmg installer


trhuxtonr:
BleachBit 6.0.3 on macOS Tahoe
Building a functional version for Apple Silicon (M4)
Complete record of the successful procedure
Python 3.8.18 · pyenv · GTK 3.24.52 · PyGObject 3.48.2 · GDK Quartz · Gettext · 61
languages · icon · ad-hoc signing

PART 1 — Preparing the project
The work was performed on the BleachBit source tree at ~/bleachbit. The version checked during the
procedure was 6.0.3, on the master branch.
cd ~/bleachbit
git branch --show-current
git log -1 --oneline
python3 -c "import bleachbit; print(bleachbit.APP_VERSION)
A dedicated Python 3.8.18 environment was created with pyenv:
pyenv virtualenv 3.8.18 bleachbit-venv
cd ~/bleachbit
echo bleachbit-venv > .python-version
pyenv version
python3 --version
which python3
The .python-version file makes pyenv automatically select the environment when entering the project directory.

PART 2 — Installing Python and GTK
dependencies
The GUI requires PyGObject and GTK. GObject Introspection was verified through Homebrew and the
required Python packages were installed in the virtual environment.
pkg-config --modversion gobject-introspection-1.0
# 1.86.0
python3 -m pip install pygobject psutil requests
Relevant final package versions:
Successfully installed pycairo-1.26.1 pygobject-3.48.2
brew list --versions gtk+3 glib gobject-introspection
# glib 2.88.3
# gobject-introspection 1.86.0_3
# gtk+3 3.24.52
GTK 3 was then verified from Python:
python3 -c "import gi; gi.require_version('Gtk', '3.0'); from gi.repository import Gtk;
print(Gtk.get_major_version(), Gtk.get_minor_version())"
# 3 24

PART 3 — Adapting GTK 3 to GDK Quartz on
macOS
The first GUI launch failed because GtkShim.py required DISPLAY or WAYLAND_DISPLAY. On macOS,
GTK uses the GDK Quartz backend rather than X11 or Wayland.
python3 -c "import gi; gi.require_version('Gdk', '3.0'); from gi.repository import Gdk; d=Gdk.Display.get_default();
print(d)"
# GdkQuartzDisplay
pkg-config --variable=targets gdk-3.0
# quartz
A backup of the original file was created before editing:
[...]

PR #2317, opened 2026-09-02
Safari cleaner for macOS

trhuxtonr:
This PR adds a new cleaner for Safari on macOS (cleaners/safari_macos.xml),
with 6 options: cookies, cache, history, site data, site preferences, and
crash reports.

Safari stores cookies in a proprietary binary format (Cookies.binarycookies),
which required writing a small parser/serializer in bleachbit/Cookie.py to
support selective (keep-list) deletion the same way the existing SQLite-based
Chrome/Firefox cookie cleaners work.

Tested on macOS Tahoe 26.2, Apple Silicon, against BleachBit's own GUI:
previewed and cleaned each option individually and combined, verified
resulting file/directory state on disk, and confirmed Safari continues to
function normally afterward.

Related to #2298.


az0:
@trhuxtonr 
**I appreciate your PRs** that make important steps towards an MVP of BleachBit on macOS---the GUI and Safari support are essential.

Instead of writing out all my feedback about the two PRs and going back and forth, **I'm planning to take both PRs into a temporary branch** to work on it before merging to master. Let me know if you have any concerns about the plan. (Alternatively, it could be simpler to break up the two big PRs into many smaller PRs, you may changes to them, and work iteratively as some changes depend on others.)

Outline of some feedback/changes (not complete feedback):

- [x] Both PRs have identical commit with different checksums
- [x] Add unit test for new code
- [x] Refactor long cookie function
- [x] Move new macOS-specific code from several existing modules to macOS-specific module (like Windows.py) that can be removed on Linux and Windows deployments
- [x] Check whether Safari is running
- [x] Resolve inconsistency in pager header between reader and writer
- [x] Remove two lines from .gitignore: one seems like personal for you, while other is redundant
- [ ] GtkShim may need changed to support CLI only, TUI, or wxPython?
- [x] Run `make pretty` on the CleanerML file
- [x] Avoid requiring new translations in the CleanerML file
- [x] Check how `Default/` covers three categories, but there are three paths to cover them instead of one path
- [x] Verify each action matches paths
- [ ] Add History.db under Container (and verify it exists)
- [ ] Add legacy Cookies not sandboxed (and verify it exists)

az0:
FYI

For testing the binary cookie reader, this swift command makes a cookie under `~/Library/Containers/com.apple.Safari/Data/Library/Cookies/Cookies.binarycookies`

```
swift -e 'import Foundation; HTTPCookieStorage.shared.setCookie(HTTPCookie(properties: [.domain: "example.com", .path: "/", .name: "test", .value: "1", .expires: Date(timeIntervalSinceNow: 3600)])!); exit(0)'
```

To test the writer, this lists cookies in that cookie jar

```
swift -e 'import Foundation; for c in HTTPCookieStorage.shared.cookies ?? [] { print("\(c.domain)\t\(c.name)=\(c.value)\t\(c.path)") }'
```

I tested the new writer with this scenario: one empty page followed by a page with real cookies, and swift correctly read the cookies from the second page

az0:
New branch with both your PRs combined, plus changes (work in progress) https://github.com/bleachbit/bleachbit/commits/macos-safari/

trhuxtonr:
> New branch with both your PRs combined, plus changes (work in progress) https://github.com/bleachbit/bleachbit/commits/macos-safari/

Thanks for the update! I've pulled macos-safari locally and reviewed the changes — really appreciate the reorganization into Mac.py, and glad the cookie parser/writer logic held up well after moving it over. No concerns about the plan; happy to wait while you continue iterating, and let me know if there's anything concrete I can help test or verify in the meantime (e.g. the pager header consistency point on your list).

trhuxtonr:
@az0 One more small thing, if useful: Safari's History cleaner didn't warn that iCloud Sync, if enabled, can re-populate deleted history from another signed-in Apple device shortly after clearing it locally — confirmed by hand (cleared history with Sync on, watched it come back; then disabled Sync and confirmed it stayed cleared, even offline).

Added a <warning> to the history option (same pattern as an existing one in claude.xml), which BleachBit already surfaces as a one-time confirmation dialog when the option is selected — so it doesn't block normal use, just informs once. Branch here if you'd like to take a look: https://github.com/trhuxtonr/bleachbit/compare/macos-safari...trhuxtonr:bleachbit:safari-icloud-history-warning

trhuxtonr:
@az0 Fixed the pager header inconsistency you flagged: _read_safari_cookie_records() and list_safari_cookies() trusted cookie_count without validating the page marker or the end-of-table marker after the offset table -- both constants the writer (and real Safari) always emit. A corrupted or hand-crafted file with a bogus cookie_count could have incorrect offsets read as cookie records instead of being rejected.

Worth flagging how I caught this: my first draft of the marker constant had the bytes wrong (b'\x00\x01\x00\x00' instead of the real b'\x00\x00\x01\x00'), which I only found by testing against my own real, unmodified Cookies.binarycookies file -- with the wrong value, the validation would have rejected every genuine Safari cookies file while accepting none. Confirmed the correct value against 10 real pages before finalizing.

Branch here if useful: https://github.com/trhuxtonr/bleachbit/compare/macos-safari...trhuxtonr:bleachbit:pager-header-fix

az0:
Merged in 173259fa3d45feb75557e7bf38720ffe7b858339 . Thank you!

PR #2319, opened 2026-09-04
Unix.py: fix get_trash_paths() leaving emptied folders behind in ~/.T…

trhuxtonr:
…rash

get_trash_paths() called children_in_directory(dirname, False) for the macOS-style flat trash, under a comment claiming this made the scan 'non-recursive'. The second parameter is actually list_directories, not a recursion toggle: children_in_directory() always recurses into subdirectories regardless of its value: with False it yields only files, descending into folders without ever yielding the folders themselves for deletion.

The practical effect: sending a folder to the Trash and then running BleachBit's 'Trash' cleaner deleted every file inside it but left the now-empty folder behind permanently, since it was never yielded as something to delete. This matches an existing docstring in FileUtilities.py: 'When list_directories is set, every directory found ... is collected into pending_dirs to be emitted after its contents' -- i.e. folders are only ever yielded when list_directories=True.

Reproduced with a real ~/.Trash folder before fixing: BleachBit reported deleting the file inside successfully, but the containing folder remained afterward, with a follow-up preview correctly showing 0 files left (nothing inside it), while the empty folder itself stayed in the Trash indefinitely.

tests/TestUnix.py adds
test_get_trash_paths_includes_now_empty_folder, using a temp directory mocked in place of ~/.Trash for a controlled, deterministic regression test; confirmed to fail against the pre-fix code (the folder path missing from the yielded paths) before being finalized.

### DIFF
--- a/bleachbit/Unix.py
+++ b/bleachbit/Unix.py
@@ -426,9 +426,12 @@ def get_trash_paths():
     # Import here to avoid a circular import.
     # pylint: disable=import-outside-toplevel
     from bleachbit import Command
-    # macOS-style flat trash (non-recursive)
+    # macOS-style flat trash. list_directories=True is required so that
+    # a folder sent to Trash is itself removed after its contents are
+    # deleted; with False, only files inside it are yielded and the now-
+    # empty folder is left behind forever.
     dirname = os.path.expanduser("~/.Trash")
-    for filename in children_in_directory(dirname, False):
+    for filename in children_in_directory(dirname, True):
         yield Command.Delete(filename)
     # Freedesktop trash spec directories
     # https://specifications.freedesktop.org/trash-spec/trashspec-1.0.html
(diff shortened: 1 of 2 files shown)


az0:
Windows raises exception :

`AttributeError: module 'os' has no attribute 'getuid'. Did you mean: 'getpid'?`

trhuxtonr:
@az0 Thanks for catching that -- fixed. The new test called get_trash_paths() directly without inheriting the @common.skipIfWindows guard that the pre-existing test_get_trash_paths already has (since the function calls os.getuid() for the Freedesktop trash spec section, which doesn't exist on Windows). Pushed the fix.

Worth mentioning: I only have macOS to test on, so os.getuid() ran fine locally and this only surfaced once it hit your Windows CI runner. Let me know if there's anything else worth double-checking given that limitation.

az0:
Merged to master in 173259fa3d45feb75557e7bf38720ffe7b858339 . Thank you!

PR #2320, opened 2026-09-04
Chrome old versions cleanup, google_chrome.xml: clean orphaned old versions left behind in the .app bundle on macOS

trhuxtonr:
Google's macOS auto-updater has never reliably removed the previous version(s) after updating — a bug publicly reported since at least 2011 (see e.g. community.jamf.com's "Google Chrome autoupdating accumulates older versions", blog.francoismaillet.com's "Is your Chrome bigger than mine?", and multiple ghacks.net articles from 2011/2012). Over an installation's lifetime, these orphaned folders under `Contents/Frameworks/Google Chrome Framework.framework/Versions/` can accumulate to several GB.

`bleachbit/Unix.py` adds `orphaned_framework_versions(versions_dir)`, which yields every version folder's contents except the one the `Current` symlink points to. Deliberately never compares version numbers or folder modification times — version numbers aren't orderable across schemes Chrome has used (e.g. `152.0.7977.83` vs `152.1.94.117`), and modification time can be misleading (confirmed by accident while testing: a folder restored from an older backup had a *newer* mtime than the genuinely-current one it sat next to). Only `Current` identifies the version actually in use; if it's missing or broken, nothing is removed.

Files inside each orphaned folder are yielded individually (folder last) rather than as one unit, since `Command.Delete` sizes a directory via a single `lstat()` — reflecting only the directory inode itself, not its actual recursive disk usage.

`bleachbit/Action.py` adds `ChromeOrphanedFrameworkVersions`, a new `macos.orphaned_framework_versions` action, no-op on non-macOS platforms.

Verified against a real Chrome installation on this machine: reinstalled an older Chrome version from Homebrew's own cache and let Chrome's normal auto-update create the leftover, then confirmed the preview correctly identified ~735MB across ~550 files as safe to remove, and confirmed a real clean successfully freed the space while leaving the current, in-use version — and Chrome itself — completely untouched.

`tests/TestUnix.py` adds `OrphanedFrameworkVersionsTestCase` (7 tests) and `tests/TestAction.py` adds a test for the new action, all confirmed to fail against the pre-fix code before being finalized.

Second commit adds the Spanish translation for the new option's label and description.

### DIFF
--- a/cleaners/google_chrome.xml
+++ b/cleaners/google_chrome.xml
@@ -251,4 +251,17 @@ later.  See the COPYING file in the top-level directory.
     <description>Settings for individual sites</description>
     <action command="json" search="file" path="$$profile$$/Preferences" address="profile/content_settings/exceptions"/>
   </option>
+  <option id="old_versions">
+    <label>Old versions</label>
+    <description>Remove old versions of Chrome left behind by updates, which Google's own updater does not clean up</description>
+    <!-- Google's macOS auto-updater has never reliably removed the
+         previous version(s) after updating, a bug publicly reported
+         since at least 2011: these can accumulate to several GB over
+         an installation's lifetime. Only the version the 'Current'
+         symlink points to is genuinely in use, so
+         orphaned_framework_versions() (bleachbit.Unix) removes every
+         sibling folder except that one; if 'Current' is missing or
+         broken, nothing is removed. -->
+    <action command="macos.orphaned_framework_versions" os="macos" search="file" path="/Applications/Google Chrome.app/Contents/Frameworks/Google Chrome Framework.framework/Versions"/>
+  </option>
 </cleaner>
--- a/cleaners/google_chrome.xml
+++ b/cleaners/google_chrome.xml
@@ -251,7 +251,7 @@ later.  See the COPYING file in the top-level directory.
     <description>Settings for individual sites</description>
     <action command="json" search="file" path="$$profile$$/Preferences" address="profile/content_settings/exceptions"/>
   </option>
-  <option id="old_versions">
+  <option id="old_versions" os="macos">
     <label>Old versions</label>
     <description>Remove old versions of Chrome left behind by updates, which Google's own updater does not clean up</description>
     <!-- Google's macOS auto-updater has never reliably removed the
--- a/tests/TestAction.py
+++ b/tests/TestAction.py
@@ -24,7 +24,7 @@ from xml.sax.saxutils import quoteattr
 
 # first party imports
 from bleachbit import IS_WINDOWS, IS_POSIX, IS_LINUX, FS_CASE_SENSITIVE, logger
-from bleachbit.Action import ActionProvider, Command, Delete, has_glob, expand_multi_var
+from bleachbit.Action import ActionProvider, ChromeOrphanedFrameworkVersions, Command, Delete, has_glob, expand_multi_var
 from bleachbit.CleanerML import CleanerML
 from tests import common
 from tests.TestFileUtilities import ini_helper
@@ -568,6 +568,40 @@ class ActionTestCase(common.BleachbitTestCase):
                     self.assertEqual(IS_LINUX, cleaner.is_usable())
                     self.assertTrue(cleaner.auto_hide())
                     mock_run_external.assert_not_called()
+
+    def test_chrome_orphaned_framework_versions_delegates_to_unix(self):
+        """ChromeOrphanedFrameworkVersions.get_commands() must delegate
+        to Unix.orphaned_framework_versions() and wrap each yielded path
+        in a Command.Delete, and must yield nothing at all on
+        non-macOS platforms (where bleachbit.Unix may not even be
+        importable)."""
+        # search="file" only registers a path get_paths() will later
+        # yield if it exists on disk, so a real (but empty) temp
+        # directory stands in for a genuine Versions/ folder.
+        versions_dir = self.mkdtemp(prefix='bleachbit-test-versions')
+        action_xml = (
+            f'<action command="macos.orphaned_framework_versions" '
+            f'search="file" path="{versions_dir}"/>')
+        action_node = parseString(action_xml).childNodes[0]
+        action = ChromeOrphanedFrameworkVersions(action_node)
+
+        fake_orphans = [f'{versions_dir}/150.0.1.1/file.txt',
+                        f'{versions_dir}/150.0.1.1']
+        with mock.patch('bleachbit.Action.IS_MAC', True), \
+                mock.patch('bleachbit.Action.Unix.orphaned_framework_versions',
+                          return_value=iter(fake_orphans)) as mock_orphans:
+            commands = list(action.get_commands())
+
+        mock_orphans.assert_called_once_with(versions_dir)
+        self.assertEqual([c.path for c in commands], fake_orphans)
+        self.assertTrue(all(isinstance(c, Command.Delete) for c in commands))
+
+        with mock.patch('bleachbit.Action.IS_MAC', False), \
+                mock.patch('bleachbit.Action.Unix.orphaned_framework_versions') as mock_orphans_win:
+            commands = list(action.get_commands())
+
+        self.assertEqual(commands, [])
+        mock_orphans_win.assert_not_called()
 
 
 def main():
--- a/tests/TestAction.py
+++ b/tests/TestAction.py
@@ -569,6 +569,7 @@ class ActionTestCase(common.BleachbitTestCase):
                     self.assertTrue(cleaner.auto_hide())
                     mock_run_external.assert_not_called()
 
+    @common.skipUnlessMac
     def test_chrome_orphaned_framework_versions_delegates_to_unix(self):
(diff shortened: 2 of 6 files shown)


tests/TestAction.py
```diff
@@ -569,6 +569,40 @@ def test_package_manager_missing(self):
                     self.assertTrue(cleaner.auto_hide())
                     mock_run_external.assert_not_called()
 
+    def test_chrome_orphaned_framework_versions_delegates_to_unix(self):
```
az0, line 573: To fix error running test on Windows, add decorator to skip the test unless macOS

cleaners/google_chrome.xml
```diff
@@ -251,4 +251,17 @@ later.  See the COPYING file in the top-level directory.
     <description>Settings for individual sites</description>
     <action command="json" search="file" path="$$profile$$/Preferences" address="profile/content_settings/exceptions"/>
   </option>
+  <option id="old_versions">
```
az0: This `os` filter hides the option on other platforms on which it would not work.

```suggestion
  <option id="old_versions" os="macos">
```

cleaners/google_chrome.xml
```diff
@@ -251,4 +251,17 @@ later.  See the COPYING file in the top-level directory.
...
+    <description>Remove old versions of Chrome left behind by updates, which Google's own updater does not clean up</description>
+    <!-- Google's macOS auto-updater has never reliably removed the
+         previous version(s) after updating, a bug publicly reported
+         since at least 2011: these can accumulate to several GB over
+         an installation's lifetime. Only the version the 'Current'
+         symlink points to is genuinely in use, so
+         orphaned_framework_versions() (bleachbit.Unix) removes every
+         sibling folder except that one; if 'Current' is missing or
+         broken, nothing is removed. -->
+    <action command="macos.orphaned_framework_versions" os="macos" search="file" path="/Applications/Google Chrome.app/Contents/Frameworks/Google Chrome Framework.framework/Versions"/>
```
az0, line 265: This doesn't block this PR, but there is an opportunity to check other paths (if applicable):

- stable Google Chrome under `~/Applications`
- non-stable Google Chrome channels: Beta, Dev, and Canary
- derived Chromium browsers

trhuxtonr: Thanks for the review! Pushed both fixes:
- Added os="macos" to the option, matching the rest of the file.
- Added @common.skipUnlessMac to the test -- both branches patch bleachbit.Action.Unix.orphaned_framework_versions, which doesn't exist as an attribute on Windows at all (Unix is only imported inside `if IS_POSIX:`), so the patch itself fails there regardless of IS_MAC's value.

On the other paths (~/Applications, Beta/Dev/Canary, Chromium derivatives) -- good point, I hadn't checked those. I'll look into it, but wanted to get these two fixes in first rather than block on it.

trhuxtonr:
@az0 Follow-up: found a real case this fix doesn't handle -- orphaned versions created by Chrome's own legitimate auto-updater are typically owned by root:wheel (confirmed via ls -la/stat), since Chrome's updater on macOS runs with elevated privileges to write into /Applications. BleachBit, running as a normal user, can't delete them: preview correctly identifies and marks the folder, but the actual delete silently fails per-file (0 files deleted, N errors), with no clear indication to the user of why.

The original verification for this PR (reinstalling an older Chrome version from Homebrew's own cache) happened to produce a folder owned by my own user, not root, so this case never surfaced there.

Two things worth considering, not proposing either as the fix yet:
- Detect and report the permission failure clearly to the user, instead of silently reporting N errors with no context.
- Whether privilege elevation (BleachBit doesn't have any macOS equivalent to Windows's elevate_privileges/UAC) is in scope for this option at all, given how much security hardening has gone into avoiding exactly that kind of surface recently.

Happy to open a fresh PR for the detection/reporting side if that seems reasonable, but wanted to flag it here first given it's the same feature.

az0:
> The original verification for this PR (reinstalling an older Chrome version from Homebrew's own cache) happened to produce a folder owned by my own user, not root, so this case never surfaced there.

Your case was an unusual exception? 
Would you guess that for >99% of macOS users, elevated privileges are required? 

> Detect and report the permission failure clearly to the user, instead of silently reporting N errors with no context.

This is a general problem that needs a general solution. On Linux, a common examples `system.localizations`, apt, and snap. On Windows, there's `system.updates` (Update installers). 

Some special functions do have systems that communicate the issue. Two examples:

- `Windows.delete_updates` checks for admin permissions and fails, even before trying to make changes
- `Unix.dnf_autoremove` executes the command. Instead of giving big traceback, it translates the error into a RuntimeError that shows as a single line.

However, `system.localizations` would act like the way you describe the Chrome old version cleanup: it could return ~100K lines of errors that aren't helpful.

If elevated privileges are always (or nearly always) required for Chrome old version cleanup, then a general solution is extending CleanerML with an option or action attribute that checks permission before running. It would show a simple, helpful error. An analog is the check for running processes (like a browser running).

If elevated privileges are often required and the action is expected to either fully work or fully fail (the way you are describing the Chrome version cleanup), then an idea: Extend CleanerML similarly with a new attribute, with the difference: the new behavior is that it tries to delete at least one file to probe whether deletion succeeds.

Some cleaning options or actions may have a mix of successes without elevated privileges (e.g., temporary directories shared between accounts), so not all can be expected to be 

> Whether privilege elevation (BleachBit doesn't have any macOS equivalent to Windows's elevate_privileges/UAC) is in scope for this option at all, given how much security hardening has gone into avoiding exactly that kind of surface recently.

Maybe power users on macOS would use this cleaning option from the CLI? Also, in the not distant future I'd like to improve privilege escalation on all platforms, but it's not simple yet.

trhuxtonr:
@az0 I’ve got it mostly solved. It was as simple as launching the app from the terminal with:

sudo open ~/carpetas-bleachbit/BleachBit-Portable/BleachBit.app

and it works. The only thing left is to implement this functionality without having to launch it with sudo. I’m working on it, and it will most likely be solved tomorrow.

az0:
>  I’ve got it mostly solved. It was as simple as launching the app from the terminal with:
> 
> sudo open ~/carpetas-bleachbit/BleachBit-Portable/BleachBit.app
> 
> and it works.

This makes sense.

> The only thing left is to implement this functionality without having to launch it with sudo. I’m working on it, and it will most likely be solved tomorrow.

What approach are you taking?

trhuxtonr:
@az0 Update: confirmed the fix -- running BleachBit with sudo (using the venv's own python3, or the packaged .app's own executable directly) successfully deletes the root-owned folder with 0 errors, so no macOS-specific limitation there.

Opened a small follow-up PR (#[número que le den a hint-permission-denied-macos]) adding a hint line to the summary when files can't be deleted due to permissions, suggesting the user try running with administrator privileges -- purely informational, doesn't attempt any elevation itself.

Separately, for a nicer UX than requiring the whole app to be launched with sudo: macOS has a standard per-action elevation pattern via osascript -e 'do shell script "..." with administrator privileges', which triggers the native password prompt (the same one OnyX and similar utilities use) for just the specific privileged operation, rather than running the entire app as root. Happy to explore that as a proper follow-up if it seems worth it, but wanted to flag the idea rather than just build it unprompted given how much security hardening has gone into this area recently.

trhuxtonr:
@az0 Replying to "What approach are you taking?" — sorry for the delay.

I want to address your earlier point too, since I don't think I responded to it directly: whether elevation is in scope here at all, given the security hardening focus. I hear that concern, and I don't think this should be decided by me building something and presenting it as done — so treat what follows as an experiment report, not a proposal to merge as-is.

I went with macOS's native per-action elevation (osascript ... with administrator privileges, the same mechanism OnyX uses) rather than launching the whole app with sudo. While testing from the actual .app (not a terminal) it consistently died with SIGKILL. Root cause, confirmed via the unified system log (cs_invalid_page, kernel code-signing check): osascript — a signed Apple binary — was inheriting DYLD_LIBRARY_PATH from BleachBit's bundled Python, and the kernel kills any strictly-signed binary that tries to honor an inherited library path like that. Fix: give osascript a clean environment. Verified with a regression test (fails without the fix, passes with it), the full suite (764 tests, no regressions), and by hand on a real double-click launch — successful deletion with the real admin prompt (screenshot attached, username blurred), declining the retry, and canceling the password prompt. All three behave safely; nothing is deleted on cancel.

Known rough edge: the system prompt shows "osascript" as the requester, not "BleachBit" — Authorization Services attributes it to the actual signed binary, not the parent app. Fixing that would need a real code signature.

On your CleanerML idea (a permission-check or probe-delete attribute): that seems like it'd compose well with either outcome here — it could be the general mechanism that detects the need for elevation across any cleaner (Chrome, system.localizations, etc.), independent of whether per-action osascript elevation, the existing sudo-the-whole-app path, or something else entirely ends up being how BleachBit actually performs the elevated delete on macOS.

Nothing from this is committed — still local, on a branch. Let me know if it's worth a proper PR, if you'd rather I focus on the CleanerML detection side instead, or if you'd prefer to hold off on elevation altogether for now.

Captura_2026-09-12_5_37_05_difuminada

PR #2324, opened 2026-09-07
bleachbit/__init__.py: fix locale directory resolution in macOS .app …

trhuxtonr:
…bundle

The packaged macOS .app bundle only ever falls back to the generic '/usr/share/locale/' path, so every bundled translation is silently lost regardless of how many languages are compiled into Contents/locale/ -- confirmed on a fresh build of current master with all 61 languages compiled and copied into place: the running app still defaulted to en/C locale, logging 'no compiled translation for language X' for every single one.

The cause: locale_dir resolution checks for a 'locale' directory next to the executable (an AppImage-style layout), but in the .app bundle bleachbit_exe_path is Contents/Resources (the parent of the bleachbit package directory), one level below the real Contents/locale directory, so that check fails and falls through to the system-wide fallback instead.

Not sure whether this is a known limitation and translations are deliberately left out of macOS builds until some later stage -- happy to close this if so. If not, this adds a macOS-specific check for the bundle-relative path, extracting the whole resolution into a small pure function, _resolve_locale_dir(), so the platform-specific branching can be tested directly.

tests/TestCommon.py adds ResolveLocaleDirTestCase with four tests, each confirmed to fail with an ImportError against the pre-fix code before being finalized.

### DIFF
--- a/tests/TestCommon.py
+++ b/tests/TestCommon.py
@@ -239,3 +239,53 @@ class CommonTestCase(common.BleachbitTestCase):
         os.remove(relative_fn)
         self.assertNotExists(relative_fn)
         self.assertNotExists('this-does-not-exist')
+
+
+class ResolveLocaleDirTestCase(common.BleachbitTestCase):
+    """Test case for bleachbit._resolve_locale_dir()"""
+
+    def test_macos_app_bundle(self):
+        """On macOS, prefer Contents/locale (one level above
+        Contents/Resources) over the generic /usr/share/locale/
+        fallback, when it exists."""
+        from bleachbit import _resolve_locale_dir
+        exe_path = '/Applications/BleachBit.app/Contents/Resources'
+        bundle_dir = '/Applications/BleachBit.app/Contents/locale'
+        result = _resolve_locale_dir(
+            exe_path, is_linux=False, is_mac=True, is_windows=False,
+            is_netbsd=False, is_bsd=False,
+            path_exists=lambda p: p == bundle_dir)
+        self.assertEqual(result, bundle_dir)
+
+    def test_macos_no_bundle_falls_back(self):
+        """On macOS, when neither the local, AppImage-style, nor
+        bundle-relative locale directory exists, fall back to the
+        generic system path."""
+        from bleachbit import _resolve_locale_dir
+        result = _resolve_locale_dir(
+            '/some/dev/path', is_linux=False, is_mac=True,
+            is_windows=False, is_netbsd=False, is_bsd=False,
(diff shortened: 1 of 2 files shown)


tests/TestCommon.py
```diff
@@ -239,3 +239,53 @@ def test_assertExists_relative_path(self):
...
+class ResolveLocaleDirTestCase(common.BleachbitTestCase):
+    """Test case for bleachbit._resolve_locale_dir()"""
+
+    def test_macos_app_bundle(self):
+        """On macOS, prefer Contents/locale (one level above
+        Contents/Resources) over the generic /usr/share/locale/
+        fallback, when it exists."""
+        from bleachbit import _resolve_locale_dir
+        exe_path = '/Applications/BleachBit.app/Contents/Resources'
+        bundle_dir = '/Applications/BleachBit.app/Contents/locale'
```
az0: Try

```suggestion
        bundle_dir = os.path.normpath(os.path.join(exe_path, '..', 'locale'))
```

There is an error 


```
_______________ ResolveLocaleDirTestCase.test_macos_app_bundle ________________
[gw2] win32 -- Python 3.12.13 D:\a\bleachbit\bleachbit\vcpkg_installed\x86-windows\tools\python3\python.exe

self = <tests.TestCommon.ResolveLocaleDirTestCase testMethod=test_macos_app_bundle>

    def test_macos_app_bundle(self):
        """On macOS, prefer Contents/locale (one level above
        Contents/Resources) over the generic /usr/share/locale/
        fallback, when it exists."""
        from bleachbit import _resolve_locale_dir
        exe_path = '/Applications/BleachBit.app/Contents/Resources'
        bundle_dir = '/Applications/BleachBit.app/Contents/locale'
        result = _resolve_locale_dir(
            exe_path, is_linux=False, is_mac=True, is_windows=False,
            is_netbsd=False, is_bsd=False,
            path_exists=lambda p: p == bundle_dir)
>       self.assertEqual(result, bundle_dir)
E       AssertionError: '/usr/share/locale/' != '/Applications/BleachBit.app/Contents/locale'
E       - /usr/share/locale/
E       + /Applications/BleachBit.app/Contents/locale

tests\TestCommon.py:258: AssertionError
```

trhuxtonr: Thanks for catching that -- fixed. The test compared against a hardcoded forward-slash literal, but _resolve_locale_dir() builds the path with os.path.normpath(os.path.join(...)), which uses backslashes on Windows. Applied your suggestion to build the expected path the same way the function does, so the comparison works regardless of platform. Pushed.

PR #2326, opened 2026-09-08
Lint fixes and tweaks

XhmikosR:


az0:
@XhmikosR 
Thank you for more fixes.

This PR is marked draft, so let me know when it's ready.

XhmikosR:
@az0: the thing is that new issues keep slipping in... CodeQL, autopep8 etc. I managed to fix/suppress all of them, but unless CI fails hard, more will end up in the code.

Also,

1. I don't like the cross-platform suppressions. They are too much noise. I will look if there's a better way.
2. The lint workflow is hitting the network and it's slow at times

az0:
> the thing is that new issues keep slipping in... CodeQL, autopep8 etc. I managed to fix/suppress all of them, but unless CI fails hard, more will end up in the code.

Sounds like the game "whack a mole." 

> 1. I don't like the cross-platform suppressions. They are too much noise. I will look if there's a better way.
> 2. The lint workflow is hitting the network and it's slow at times

**Do you want to submit a PR with the no-regret lint fixes**, while delaying the cross-platform suppressions (that you want to improve) and the lint workflow (can be slow; maybe too sensitive)? (Or we can wait---ot's up to you.)

Examples of commits that could be prioritized while you figure out the rest.

0ae20d428472545722ccb40fb5fe1514e4cf5df3
9f5df55e7e002131449a76aa83dce70b508dc659
73e5c45749ce8c994c93c8a33c67cb1922e00a8b
85afd052612afaf4ce0c26938209c1a9fbb333a9

XhmikosR:
@az0: i think this is now as good as it gets, but please review the patches carefully. It might not be perfect, but it should be a good starting point to make sure quality is enforced.

CI will fail now if there are issues, which I assume you want to do so in order to keep quality. CodeQL issues depend on the severity. Direct merges to branches will only catch issues when he jobs run after the merge, so you'll still end up with new issues if you don't follow the PR workflow in the repo.

Also, might be worth enabling always suggest updating branches in the repo settings.

PR #2333, opened 2026-09-09
Mac.py: read AppleLocale from .GlobalPreferences.plist instead of she…

trhuxtonr:
…lling out

get_macos_locale() called `defaults read -g AppleLocale` as a subprocess -- reliable when the app was launched from a terminal (or via `open`, even without any arguments), but confirmed to consistently fail with 'no default locale found. Assuming C' when the exact same .app bundle was launched via a genuine Finder double-click instead, with no other change: same files, same bleachbit.ini, same machine, only the launch method differed.

Manually forcing a specific language always worked correctly regardless of launch method, isolating the failure to this one function specifically, not to translation loading in general.

Reading ~/Library/Preferences/.GlobalPreferences.plist directly via plistlib is pure file I/O with no subprocess involved, so it cannot be affected by whatever session/TCC context differs between a terminal-descended process and a genuine Finder-launched one. Confirmed by hand: after applying this fix to the running .app, auto-detect correctly resolved to es_ES on every subsequent real Finder double-click, including immediately after toggling 'Auto-detect language' back on in Preferences without even restarting the app.

The existing `defaults read -g AppleLocale` call is kept as a fallback for the (currently unobserved, but plausible) case where the plist itself cannot be read, with AppleLanguages as a further fallback if AppleLocale specifically is absent from the plist.

tests/TestMac.py replaces test_get_macos_locale_mocked (which directly mocked subprocess.run, no longer reachable in the common case now that the plist read succeeds first on any real machine) with four tests covering the plist-success path, the AppleLanguages fallback, the defaults-command fallback when the plist is unreadable, and the all-sources-fail case.

### DIFF
--- a/bleachbit/Mac.py
+++ b/bleachbit/Mac.py
@@ -12,6 +12,7 @@ import errno
 import logging
 import os
 import platform
+import plistlib
 import subprocess
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


bleachbit/Mac.py
```diff
@@ -79,6 +94,37 @@ def get_macos_locale():
...
+    On macOS, locale.getlocale() reflects POSIX environment variables
+    (LANG/LC_ALL), which are not set when the app is launched from
+    Finder (as opposed to a terminal) and can be stale or inconsistent
+    with the user's actual System Settings > General > Language & Region
+    preference. AppleLocale reflects the real system preference
+    regardless of how the app was launched.
+
+    Prefer reading it directly from .GlobalPreferences.plist (pure file
+    I/O, no subprocess); fall back to `defaults read -g AppleLocale`,
+    and finally to the first entry of AppleLanguages, if that fails.
```
az0: Change the order described in the docstring to match the code

```suggestion
       Prefer reading AppleLocale directly from .GlobalPreferences.plist
       (pure file I/O, no subprocess); if AppleLocale is absent from the
       plist, fall back to the first entry of AppleLanguages in the same
       plist, and finally to `defaults read -g AppleLocale` if the plist
       itself cannot be read.
```

bleachbit/Mac.py
```diff
@@ -57,16 +58,30 @@
...
-    """Return the user's preferred locale on macOS, e.g. 'es_ES'.
+def _read_global_preferences_plist():
+    """Return the parsed contents of ~/Library/Preferences/
+    .GlobalPreferences.plist, or None if it cannot be read."""
+    path = os.path.expanduser(
+        '~/Library/Preferences/.GlobalPreferences.plist')
+    try:
+        with open(path, 'rb') as f:
+            return plistlib.load(f)
+    except (OSError, ValueError, plistlib.InvalidFileException) as e:
```
az0:
```suggestion
    except (OSError, ValueError, plistlib.InvalidFileException,  xml.parsers.expat.ExpatError) as e:
```

bleachbit/Mac.py
```diff
@@ -12,6 +12,7 @@
 import logging
 import os
 import platform
+import plistlib
 import subprocess
```
az0, line 16:
```suggestion
import subprocess
import xml.parsers.expat
```

trhuxtonr: "Applied all three suggestions exactly as proposed in commit 05e86c8e. Thanks for the review!"

az0:
Merged in 31a1ef095efd70f797f2a425fee2c59e8b27c46a

PR #2335, opened 2026-09-10
Fix hamburger menu not following language changes on macOS

trhuxtonr:
Two separate bugs combined to freeze the hamburger menu (app-menu.ui, loaded via Gtk.Builder) in whatever language was active at process launch, regardless of any later language change made in Preferences, with or without restarting the app:

1. GuiWindow.py: create_headerbar() loaded app-menu.ui into the menu button exactly once, at initial window construction. Unlike the rest of the headerbar (whose labels/tooltips are re-set through Python's own gettext _() calls on every update_headerbar_labels()), Gtk.Builder translates a .ui file's own translatable strings once, at load time, with no later re-evaluation -- so the menu stayed permanently frozen in whichever language was active the one time it was ever built.

   Extracts the loading logic into _reload_app_menu(), storing the button as self.menu_button, and calls it again from update_headerbar_labels() (already invoked on every language change) so the menu is rebuilt, not just its surrounding labels.

2. Language.py: even with the menu correctly reloading, it still stayed in Spanish through three different manually-selected languages and a full app restart, on a machine whose real macOS System Settings > Language is Spanish. GLib's own g_get_language_names() (which Gtk.Builder uses for a .ui file's translations) reads the LANGUAGE/LC_ALL/LC_MESSAGES/LANG environment variables directly -- it does not consult locale.setlocale()'s C locale state at all, which is the only thing setup_translation() was setting on POSIX. Without LANGUAGE ever actually being set in the environment, GLib fell back to whatever the real OS environment had at launch (the real AppleLocale preference), completely bypassing BleachBit's own in-app language selection.

   setup_translation() now also sets os.environ['LANGUAGE'] on POSIX (mirroring the existing Windows-only os.environ['LANG'] line). Deliberately only LANGUAGE, not LANG/LC_ALL: those are read by every other locale-aware consumer in the process, including any subprocess spawned afterward, and a bare code without an explicit encoding (required for GLib's own parsing of these variables specifically -- confirmed by hand that a '.UTF-8' suffix breaks it) is not a well-formed LC_ALL/LANG value elsewhere, and setting them this way was confirmed by hand to crash a subprocess Python interpreter started later with 'Fatal Python error: config_get_locale_encoding: ... nl_langinfo(CODESET) failed'.

Confirmed by hand on the real .app via a genuine Finder double-click: before, the menu stayed in Spanish through English, French, and German selections and a full restart; after, it correctly switches immediately on every language change, without even restarting the app.

tests/TestLanguage.py adds SetupTranslationEnvironTestCase with two tests, confirmed to fail against the pre-fix code before being finalized. Both mock gettext.translation() and locale.setlocale() (and directly snapshot/restore the module-level Language.t): the real setup_translation() has these two further real side effects beyond the one each test is targeting, and leaving either unmocked was confirmed by hand to leak a non-English translation into unrelated later tests' own log messages for the rest of the test process.

### DIFF
--- a/bleachbit/GuiWindow.py
+++ b/bleachbit/GuiWindow.py
@@ -1023,6 +1023,13 @@ class GUI(Gtk.ApplicationWindow):
 
     def update_headerbar_labels(self):
         """Update the labels and tooltips in the headerbar buttons"""
+        # The hamburger menu is not just a label -- it's a whole
+        # Gtk.Builder-loaded Gio.MenuModel that only picks up a
+        # language change if explicitly reloaded (see the comment in
+        # _reload_app_menu()).
+        if hasattr(self, 'menu_button'):
+            self._reload_app_menu()
+
         # Preview button
         self.preview_button.set_label(PREVIEW_MSG)
         self.preview_button.set_tooltip_text(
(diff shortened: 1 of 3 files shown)


az0:
Looks good, thank you.

## Notes

I wrote notes for a follow up change for Windows

- libintl is from GNU gettext,not GLib
- intl-8.dll caches the loaded .mo per domain. Changing LANG/LANGUAGE or re-calling bindtextdomain() does not reload it. 
- Gtk.Builder uses g_dgettext (GLib API to libintl)
- The hamburger menu stays stuck after the first non-English load. 
- GNU gettext `_nl_msg_cat_cntr` can clear

Issue #1800, opened 2025-03-24
Fix auto-detect language field appearing blank and label state mismatch in preferences


banaagravrommel:
**Describe the bug**
For the aut-detect language setting, the field is blank which is wrong, it should show the auto detected language but should be disabled. Another thing when you disable/enable the checkbox the field is shown as enabled but the label is disabled.

**To Reproduce**
1. Download https://download.bleachbit.org/beta/4.9.1/BleachBit-4.9.1.2886-setup.exe
2. Install it
3. Go to preferences

The error happens (sometimes/every time).

**Expected behavior**
- The auto detected language E.G is "ENGLISH" then the field should show "ENGLISH"
- Field should be disabled same to the label
- 
**Screenshots**
1
![Image](https://github.com/user-attachments/assets/158670b2-8f7d-45d3-8e3f-857430fd91c1)
2

![Image](https://github.com/user-attachments/assets/0d5ee57a-a3d2-4699-9d97-80f3e0c5d48e)

**Desktop (please complete the following information):**
 - Windows 11
 - BleachBit version 4.9.1
 - 
**Additional context**
This is more of an improvement not a bug

Issue #1799, opened 2025-03-24
Auto detected language should be the default language used upon unchecking the auto-detect language setting


banaagravrommel:
**Describe the bug**
Upon installing Bleachbit Alpha version, the default language upon unchecking "Auto-detect language" is like this symbol language. Which should be the default language detected previously by the program.

**To Reproduce**
1. Download https://download.bleachbit.org/beta/4.9.1/BleachBit-4.9.1.2886-setup.exe
2. Install it
3. Go to preferences
4. Uncheck Auto-detect language

The error happens (sometimes/every time).
Everytime

**Expected behavior**
If the detected language is E.G "ENGLISH" then upon unchecking the Auto-detect language checkbox it should automatically select the default language.

**Screenshots**

![Image](https://github.com/user-attachments/assets/9ffe5d7b-4dee-44af-b1c3-8fc926e431d3)

**Desktop (please complete the following information):**
 - Windows 11
 - BleachBit version 4.9.1
 
**Additional context**
This is more of an improvement not a bug

PR #2336, opened 2026-09-10
GuiPreferences.py: fix language combo box showing nothing selected

trhuxtonr:
The language dropdown in Preferences compared each supported language's code against the active language code with strict string equality. BleachBit's own supported-language list uses bare base codes ('es', 'fr'), but get_active_language_code() can return a region-specific code ('es_ES') -- notably on macOS, where AppleLocale-based auto-detection (get_macos_locale()) always returns one. 'es' == 'es_ES' is false, so no entry ever matched and the combo box was left showing nothing selected, even though the app itself was correctly running in that language.

Falls back to comparing against the base code (the part before the first underscore) when no exact match is found, so a region-specific active language still highlights its corresponding entry.

Confirmed by hand on the real .app via a genuine Finder double-click: the box previously showed nothing with 'Auto-detect language' checked and es_ES active; now shows 'Español (es)' correctly selected.

tests/TestGUI.py adds test_preferences_language_combo_matches_region_specific_code, confirmed to fail (no active text in the combo) against the pre-fix code before being finalized.

### DIFF
--- a/bleachbit/GuiPreferences.py
+++ b/bleachbit/GuiPreferences.py
@@ -362,12 +362,22 @@ class PreferencesDialog:
             lang_items = [('en_us', 'English')]
         else:
             lang_items = supported_langs.items()
+        # current_lang_code may be a region-specific code like 'es_ES'
+        # (e.g. from macOS AppleLocale), while lang_items' keys are the
+        # bare base codes BleachBit ships translations under (e.g.
+        # 'es') -- matching only the base part before the underscore
+        # keeps the combo box in sync with the language actually in
+        # use instead of leaving nothing selected.
+        current_lang_base = current_lang_code.split(
+            '_')[0] if current_lang_code else None
         for lang_idx, (lang_code, native) in enumerate(lang_items):
             if native:
                 self.lang_combo.append_text(f"{native} ({lang_code})")
             else:
                 self.lang_combo.append_text(lang_code)
             if lang_code == current_lang_code:
+                active_language_idx = lang_idx
+            elif active_language_idx is None and lang_code == current_lang_base:
                 active_language_idx = lang_idx
         if active_language_idx is not None:
             self.lang_combo.set_active(active_language_idx)
--- a/tests/TestGUI.py
+++ b/tests/TestGUI.py
@@ -238,6 +238,23 @@ class GUITestCase(common.BleachbitTestCase):
         self.assertEqual(mock_get.call_count, 1)
         pref.dialog.destroy()
 
+    def test_preferences_language_combo_matches_region_specific_code(self):
+        """Regression test: when the active language code is
+        region-specific (e.g. 'es_ES', as returned by macOS's
+        AppleLocale-based auto-detection), the language combo box must
+        still show it as selected, matching against the base code
+        ('es') that BleachBit's own supported-language list uses --
+        not leave nothing selected because 'es_ES' != 'es'."""
+        with mock.patch('bleachbit.GuiPreferences.get_active_language_code',
+                        return_value='es_ES'):
+            pref = self.app.get_preferences_dialog()
+        try:
+            active_text = pref.lang_combo.get_active_text()
+            self.assertIsNotNone(active_text)
+            self.assertIn('(es)', active_text)
+        finally:
+            pref.dialog.destroy()
+
     def test_preferences_cookies_page(self):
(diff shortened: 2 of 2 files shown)


az0:
Thank you. Based on the PR description, we may be both solving the same issues. Earlier today, I wrote but haven't yet pushed a commit to fix https://github.com/bleachbit/bleachbit/issues/1800 , https://github.com/bleachbit/bleachbit/issues/1799 , and related issues. On Linux, a problem was auto-detection in `getlocale` was poisoned by `setlocale` to override it.

After I push it, perhaps you can check whether it solves your issue and whether I missed anything.

trhuxtonr:
@az0  Captura de pantalla 2026-09-11 a las 5 17 46

az0:
What do you want me to see in the screenshot?

For comparison, this is a screenshot from code running on Linux in the same language as yours, also with auto-detection enabled.

image

az0:
FYI, this is my recent attempt for fixing for language selection, but I am working through a spiderweb of corner cases. It needs more work.

https://github.com/bleachbit/bleachbit/commit/9d54144398a1a723a316ece83f61cae49752954f
https://github.com/bleachbit/bleachbit/commit/67c6cfff2a5ebbd6c76ee1362ae850c14211923e

trhuxtonr:
@az0 Since this PR and your locale poisoning finding on Linux (getlocale poisoned by setlocale) are both about locale-matching correctness, this might be relevant: found a related but distinct bug in find_best_locale() (Unix.py) while testing on macOS.

Repro: find_best_locale('es_ES') returned the bare 'es_ES' instead of 'es_ES.UTF-8', even though both exist in locale -a's output. Root cause: the exact-match check (if user_locale in available_locales: return user_locale) runs before the UTF-8-preferring search just below it, so a bare match with no encoding suffix wins first.

Effect: passing that bare locale to locale.setlocale(LC_ALL, ...) on macOS doesn't raise or fall back to UTF-8 -- it silently leaves the process's C-level codeset as ISO8859-1. That then corrupts any multi-byte UTF-8 character when GLib/libintl re-encodes a translated .mo file for a Gtk.Builder-loaded menu (the hamburger menu, translated via LANGUAGE/libintl, a separate path from Python's own gettext.translation()). Confirmed by hand: exactly the menu entries containing an accented character ("Purgar espacio vacío", "Información del sistema") showed literal [Invalid UTF-8], while every accent-free entry rendered fine -- pointed straight at an encoding-specific corruption rather than a general translation failure.

Fix: only let the exact-match check win when the caller explicitly asked for a specific encoding (user_locale itself already contains a dot); otherwise fall through to the UTF-8-preferring search, for the same reason as the existing current-locale shortcut just above it. Added test_find_best_locale_prefers_utf8_over_bare_exact_match, confirmed to fail against the pre-fix code ('es_ES' != 'es_ES.UTF-8') before finalizing. Full suite passes with no regressions.

Currently sitting locally, unpublished -- happy to open it as its own small PR if useful, or fold it into whatever you're doing here, whichever is less disruptive to your corner-case work.

az0:
> Repro: find_best_locale('es_ES') returned the bare 'es_ES' instead of 'es_ES.UTF-8', even though both exist in locale -a's output. Root cause: the exact-match check (if user_locale in available_locales: return user_locale) runs before the UTF-8-preferring search just below it, so a bare match with no encoding suffix wins first.

8f1fc1a128eb48bc356fb7cf9e9bfa8bf8030b23 should have fixed this, so would you please test with master (3796a6d40cc8502eb49def6690b6ee9b03f6670e) or at least with 01b0c97d24ed0efb30af049adf6b82d50ba3dab6 ? 

Check whether the new `test_find_best_locale_prefers_utf8_over_bare` covers what you mentioned.

trhuxtonr:
@az0 Confirmed: tested find_best_locale('es_ES') against master's current Unix.py (with 8f1fc1a1) -- now correctly returns 'es_ES.UTF-8', and locale.setlocale() leaves the process at ('es_ES', 'UTF-8') instead of the broken ISO8859-1. Your fix covers the case I reported.

I'll drop my own local fix for this since yours is already in and handles more cases (the @euro-style modifiers too). Thanks for catching it independently before I even got to report it.

az0:
Please check the new master that has 31b4dfd3070b708db660c41ca84c92aa35d3eb73 to fix this.

PR #2345, opened 2026-09-21
Further performance tweaks

XhmikosR:
@az0: please check the patches carefully. I'm not 100% sure about all of them, although tests are green.

Regarding the GUI patches, now it's almost instant to the point that scrolling might feel unnatural. But I'm not sure how to do this better, open to any suggestions.


XhmikosR:
I made significant progress regarding loading performance in https://github.com/XhmikosR/bleachbit/compare/xmr/perf...XhmikosR:bleachbit:xmr/perf2

I want to test those further (I'm testing on my Windows machines) and startup time has been reduced by about ~600ms. I will push those patches here later if they are OK so that we land them in one PR.

XhmikosR:
Specifically, [this patch](https://github.com/XhmikosR/bleachbit/commit/db9b379e500c71f713726f81e17fdcbc5eec6f93) seems to save the most time on startup, at least on Windows. Unsure if there's any side effect. 

The library.zip compression also affects load time, but the final installer and zip are smaller. The installed size grows by 5MB.

Still testing the patches from the above branch.

az0:
> please check the patches carefully. I'm not 100% sure about all of them, although tests are green.

I'm looking forward to reviewing them.

> Still testing the patches from the above branch.

The PR is marked draft, and I see that you are still actively working on it. Should I hold review for you to mark the PR ready?  Or would it help to have feedback now?

> I want to test those further (I'm testing on my Windows machines) and startup time has been reduced by about ~600ms. 

Great, 600ms is significant. I assume that's with cold cache, so the warm cache scenario could have a larger improvement.

> I will push those patches here later if they are OK so that we land them in one PR.

As a general rule, if we could please separate any high-risk or high-complexity commits, that would be appreciated. I'd like to start merging the low risk commits for release 6.0.6, which I'd like to release soon.

> Specifically, [this patch](https://github.com/XhmikosR/bleachbit/commit/db9b379e500c71f713726f81e17fdcbc5eec6f93) seems to save the most time on startup, at least on Windows. Unsure if there's any side effect.

Good job finding bottlenecks in the splash screen (which is ironic) and the library.zip compression. Neither were areas I expected to find performance improvements, and my intuition was that shrinking library.zip with more compression would help with cold-cache startup.

Have you tried the py2exe setting `skip_archive: True` ?

> Regarding the GUI patches, now it's almost instant to the point that scrolling might feel unnatural.

My intuition for "Queue one log scroll at a time instead of one per line" is that it would stutter, so I am looking forward to trying that.

XhmikosR:
> The PR is marked draft, and I see that you are still actively working on it. Should I hold review for you to mark the PR ready? Or would it help to have feedback now?

Give me a little more time to merge the patches from the other branch here too.

> Great, 600ms is significant. I assume that's with cold cache, so the warm cache scenario could have a larger improvement.

Cache doesn't have any effect. It was the keyboard hook that blocked the main thread from aprox. 600ms. I didn't see this slowness on non-Windows, and I only tested this on Windows.

> As a general rule, if we could please separate any high-risk or high-complexity commits, that would be appreciated. I'd like to start merging the low risk commits for release 6.0.6, which I'd like to release soon.

The thing is that most of these can have an impact and were found by using Claude. I verified them as good as I could but still, the more eyes the better. I'll see if I could split the patches in two PRs.

> Good job finding bottlenecks in the splash screen (which is ironic) and the library.zip compression. Neither were areas I expected to find performance improvements, and my intuition was that shrinking library.zip with more compression would help with cold-cache startup.
> 
> Have you tried the py2exe setting `skip_archive: True` ?

The thing is that compression and decompression (even if it's fast nowadays) it's still pays some CPU time. I think at this point we are better off not using library.zip at all later. I didn't try the option, but without compression it saved aprox. 30ms from loading time on my Windows 10 VM. I plan to experiment with this further in the future.

> My intuition for "Queue one log scroll at a time instead of one per line" is that it would stutter, so I am looking forward to trying that.

No stuttering at all, but again, my machines are pretty fast. You can try the builds from my fork already if you want.

XhmikosR:
BTW commit https://github.com/bleachbit/bleachbit/pull/2345/commits/2f75a00cb34a1d64294926e6a01313c2d5f6ed82 fixes a crash. You can cherry pick that separately if you prefer.

XhmikosR:
I split the fixes to #2349. I merged all perf-related patches to this branch. I only tested these patches mostly on Windows and Fedora.

The CI failure again looks unrelated, we really need to fix the flaky tests at some point. BTW I can't restart CI myself.

bleachbit/Process.py
```diff
@@ -170,12 +171,41 @@ def _enumerate_ps_aux():
...
+class ProcessCache:
+
+    """Cached way to determine whether a process is running"""
+
+    def __init__(self, max_age_seconds=10):
+        self.max_age_seconds = max_age_seconds
+        self.last_scan_time = None
+        self.processes = ()
+
+    def invalidate(self):
```
az0: Scenario: user tries to clean a browser, but BleachBit complains the browser is running. User closes it and tries again, but the cache prevents it, so user is confused.

Solution: please invalidate at start of each clean run, even if within ten seconds. Alternatively, recheck the Firefox PIDs found earlier are still alive---which I presume is faster than re-enumerating all processes.

bleachbit/GuiWindow.py
```diff
@@ -1255,9 +1323,32 @@ def check_orphaned_wipe_files(self):
...
+
+        threading.Thread(target=_worker, daemon=True).start()
+        return False
+
+    def _prompt_orphaned_wipe_files(self, orphaned_files):
+        """Ask whether to preview orphaned wipe files, on the main thread"""
+        if self.textbuffer is None:
+            # window was destroyed while the scan was running
+            return False
+        if not self.get_sensitive():
```
az0: The guard never fires, and I suggest this instead

```suggestion
        if not self.run_button_get_sensitive():
```

bleachbit/Process.py
XhmikosR: I went with the first solution and also addressed your other comment and added a test.

